PR #96, opened 2018-04-23
[DO NOT MERGE] Lootbox problem

gromakovsky:
See why we can't use the `master` branch from `lootbox`.


kirelagin:
It seems unlikely that any progress on this will happen upstream any time soon. Given that Disciplina now depends on Lootbox for logging, we can no longer wait for this, so we went ahead and applied a fix to our fork of nixpkgs.

Please, feel free to return the dependency on `master` of Lootbox, it should work now.

gromakovsky:
@kirelagin something doesn't work :man_shrugging:

kirelagin:
🤷‍♂️ I swear, before writing to you I built ariadne with lootbox master. I will look into this.

kirelagin:
Oh well 🤷‍♂️ 
```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for loot-crypto-0.0.0.0:
    base-noprelude must match -any, but the stack configuration has no specified version (latest matching version is 4.11.1.0)
needed due to ariadne-0.2.0.0 -> loot-crypto-0.0.0.0

In the dependencies for loot-prelude-0.0.0.0:
    base-noprelude must match -any, but the stack configuration has no specified version (latest matching version is 4.11.1.0)
needed due to ariadne-0.2.0.0 -> loot-prelude-0.0.0.0
```

gromakovsky:
@kirelagin approve pls.

PR #111, opened 2018-05-02
[AD-61] Make REPL widget omnipresent

goodsoft:
Much disputable very don't merge wow


gromakovsky:
We decided to make this widget omnipresent (btw, not really, I don't think it should be in `About`) and only show the last command in REPL output (to make it smaller).

goodsoft:
Not anymore, we discussed it with Gromak and Rukin and decided it's OK.

gromakovsky:
• `stack exec ariadne`
• `Esc-L`
• `Esc-R`

After that I am in `Wallet`. Is it intended?

goodsoft:
Probably not, will fix.

PR #137, opened 2018-05-20
[AD-109] Extract TUI/Qt into separate Cabal files

maksbotan:
This will allow Windows CI to build only ariadne-qt with stack,
as `vty` library can't be built on Windows.


yegortimoshenko:
@yorickvP Sounds good to me.

maksbotan:
Note that TUI is now separate as well, so you probably need to build all three packages: `ariadne`, `ariadne-vty`, `ariadne-qt`.

gromakovsky:
@yorickvP @yegortimoshenko can you do it please? :pray:

gromakovsky:
I'll re-review when CI passes.

PR #148, opened 2018-05-25
[AD-118] Add address checkpoints

Pastafarianist:
There are a few disputable decisions made here, please review with care.

* Account checkpoints remain in place, only the type `Checkpoint` is renamed to `AccCheckpoint`.
* Address checkpoints are added. They contain address balance.
* All state-modifying functions (`applyBlock`, `newPending`, `switchToFork`) are now generalized to work on `UpdatableWalletState`. There are two instances of it (both are stubbed with `error`): one for `AccCheckpoints` and the other for `AddrCheckpoints`.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
+++ b/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
@@ -85,10 +85,12 @@ deriveSafeCopySimple 1 'base ''NewPendingError
 newPending :: HdAccountId
            -> InDb (Core.TxAux)
            -> Update DB (Either NewPendingError ())
-newPending accountId tx = runUpdate' . zoom dbHdWallets $
+newPending accountId tx = runUpdate' . zoom dbHdWallets $ do
     zoomHdAccountId NewPendingUnknown accountId $
-    zoom hdAccountCheckpoints $
-      mapUpdateErrors NewPendingFailed $ Spec.newPending tx
+      zoom hdAccountCheckpoints $
+        mapUpdateErrors NewPendingFailed $ Spec.newPending tx
+    -- TODO: update address checkpoints for child addresses
+    -- of this account
 
 -- | Apply a block
 --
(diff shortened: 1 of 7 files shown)


ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -89,8 +90,22 @@ newPending accountId tx = runUpdate' . zoom dbHdWallets $ do
...
+        coerceAction addr $ mapUpdateErrors NewPendingFailed $ Spec.newPending tx
+      put newAddresses
+  where
+    coerceAction
+        :: HdAddress
+        -> Update' AddrCheckpoints NewPendingError ()
+        -> Update' a NewPendingError HdAddress
+    coerceAction addr action =
+        let oldCheckpoints = view hdAddressCheckpoints addr
+            checkpointsOrErr = runIdentity $ runExceptT $ runStateT action oldCheckpoints
```
gromakovsky: `runIdentity $ runExceptT` → `runExcept $`

gromakovsky: Also consider using `execStateT` instead of `runStateT`.

ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet/Read.hs
```diff
@@ -72,12 +72,11 @@ check f g = using' f (const g)
...
-    sumCoins
-    . map hdAddressBalance
-    . toList
-    . IxSet.getEQ rootId
-    . view hdWalletsAddresses
+hdRootBalance rootId = sumCoins
+                     . map hdAccountBalance
+                     . toList
+                     . IxSet.getEQ rootId
+                     . view hdWalletsAccounts
```
gromakovsky: It's better to squash this commit with the one where it was introduced.

Pastafarianist: I will squash everything into a single commit when I merge.

gromakovsky: Well, anyway, it's better to squash before requesting review, because some people review per commit.

PR #166, opened 2018-05-29
[AD-132] Rename number types in help

maksbotan:
Specify bounds for number types more explicitly:
Word8 -> 8-bit UInt.

---

I did not touch `Int` as bounds for it are not fixed in Haskell, according to docs:

> A fixed-precision integer type with at least the range [-2^29 .. 2^29-1].

Also, I capitalized `UInt` to match our other types.


int-index:
I think there's an assumption in the pretty printer that type names are a single lexeme, so they are never parenthesized. Better use `UInt32` instead of `32-bit UInt` etc

PR #190, opened 2018-06-13
[AD-204] Add git commit hash to the version message

hithroc:
Tested with regular build and `stack build --nix`


ariadne/src/Ariadne/Config/TH.hs
```diff
@@ -0,0 +1,20 @@
...
+import Formatting (sformat, string, (%))
+import Language.Haskell.TH (ExpQ, litE, stringL, runIO)
+import System.Process (readProcess)
+
+getCommitHash :: ExpQ
+getCommitHash = do
+  hash <- runIO $
+    -- `tryAny` here will both catch exit code and other exceptions like bad file descriptor
+    -- which can occur if git executable was not found, for example
+    tryAny (readProcess "git" ["rev-parse", "--verify", "HEAD"] "") >>= \case
```
gromakovsky: I'd use `strip`, because IIRC otherwise it will have trailing newline and I don't like it (I expect `getCommitHash` return only hash, without any newlines).

hithroc: There's no `strip` for `String`, so I made my own. It's not exactly graceful, but good enough here I think.

gromakovsky: I'd convert to `Text` and use `Text.strip`, but here it's not essential indeed.

PR #196, opened 2018-06-18
[AD-17] Switch Command History to be backed by sqlite

mitchellvitez:
One issue: if a file called `ariadne_history` already exists, we get an error when trying to create the DB


ariadne/src/Ariadne/UX/CommandHistory.hs
```diff
@@ -27,7 +31,8 @@ openCommandHistory historyFile = do
     counter <- newIORef (-1)
     currentPrefix <- newIORef ""
 
-    unlessM (doesFileExist historyFile) $ writeFile historyFile ""
+    conn <- open historyFile
```
gromakovsky: [The documentation](https://www.stackage.org/haddock/lts-11.2/sqlite-simple-0.4.14.0/Database-SQLite-Simple.html#v:open) states that every `open` must be closed with a call to `close`, but it's not the case here.
I'd use `withConnection` instead.
The same applies to two other places where `open` is used.

gromakovsky:
Btw, `ariadne_history.db` should be added to `.gitignore` (and I would keep `ariadne_history` there for a while and delete it later).

mitchellvitez:
Still working on an issue where the first result in the history will be ignored if going back and forth through history

goodsoft:
You are still changing `newCounter` by `+1` or `-1`, which is wrong, as ids may not go one after another.
This is especially true in case there's prefix, as only some of the commands in history will match, and this counter will need to skip several ids.

Also, your problem is that you initialize this counter by last command id in history (say, `10` for 10 commands), and when you try to get previous command for the first time, you _at first_ decrement this counter (so you have `9`), and only _afterwards_ get a row from db with that is `<=` that id, so you _always_ skip the last command.

I think you should handle it this way:
* Initialize `currentHistoryItem` or something with `select max(id) + 1 ...`
* Use `>` and `<` in queries
* Instead of incrementing or decrementing `currentHistoryItem` just set it to the `id` of the row you got from db
* `counterChange` should be of enum type, like `Back`/`Forward` or `LT`/`GT`, as we don't have a need for its numerical value, just the direction

goodsoft:
Also, I think that in case of empty response from db the behavior should differ, depending on whether we reached the beginning or the end of history.

In the former case we shouldn't do anything, i.e. our cursor stays at the earliest matching history item and doesn't move. Thus, we will be able to hold down `↑` key to scroll until the very beginning and afterwards press `↓` to traverse history in the other direction without resetting cursor.

In the latter case we just reset it to `max(id) + 1` and show our originally typed prefix.

mitchellvitez:
With the new behavior on edges, we're always returning a `Just ...` in `changeCommand`. Wasn't sure whether we wanted to change the signature to `IO Text` instead of `IO (Maybe Text)`, because there may be a use for `Nothing` there in the future

PR #197, opened 2018-06-19
[AD-233] Add "account" arguments to the 'send' command

gromakovsky:
Only interface is provided, because it's needed for UI.
It will be implemented a bit later, after we finally start using
the new backend.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Face.hs
+++ b/ariadne/src/Ariadne/Wallet/Face.hs
@@ -4,6 +4,7 @@ module Ariadne.Wallet.Face
   , WalletEvent(..)
   , WalletReference(..)
   , AccountReference(..)
+  , LocalAccountReference(..)
   , WalletSelection(..)
   , WalletName(..)
   , Mnemonic(..)
@@ -36,6 +37,13 @@ data AccountReference
   | AccountRefByIndex !Word32 !WalletReference
   | AccountRefByName !Text !WalletReference
 
+-- | Reference to an account inside a wallet.
+data LocalAccountReference
+  = LocalAccountRefByIndex !Word
+  -- ^ Reference by index in UI.
+  | LocalAccountRefByName !Text
+  -- ^ Reference by account name.
+
 -- | Single string representing a mnemonic, presumably space-separated
 -- list of words.
 newtype Mnemonic = Mnemonic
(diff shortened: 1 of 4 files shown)


docs/usage-tui.md
```diff
@@ -198,8 +201,20 @@ So, to construct an output, do this:
...
+index (`wallet` argument) as well as list of source accounts (as names
+or indices). If you don't specify a wallet, the selected one will be used.
+
+Transaction inputs are selected automatically from a set of accounts
+which depends on `account` arguments and current selection:
+* If at least one account is explicitly specified, only the specified
+  accounts can be used as inputs.
+* If no accounts are explicitly specified and an account from the
+  input wallet is selected, only its addresses will be used as inputs.
+* Otherwise, addresses will be picked from all accounts in the input
```
maksbotan: This seems to contradict help, right?

> If an account in the input wallet is selected, only this account will be used as input.

docs/usage-tui.md
```diff
@@ -73,7 +73,10 @@ and these functions in more detail.
 ### General syntax
 
 To begin with, all `Knit` functions accept keyword arguments, some of which may be optional.
-Optional arguments are marked with "`?`" suffix in the help. To specify a keyword argument, write
+Optional arguments are marked with "`?`" suffix in the help. If an
+argument can be specified many times (including zero), it's marked
+with "`*`" suffix. If it can be specified many times, but at least
+once, it's marked with "`+`" suffix. To specify a keyword argument, write
```
maksbotan: I wanted to avoid this regex BS in an _user_ manual :unamused:

ariadne/src/Ariadne/Wallet/Face.hs
```diff
@@ -36,6 +37,13 @@ data AccountReference
   | AccountRefByIndex !Word32 !WalletReference
   | AccountRefByName !Text !WalletReference
 
+-- | Reference to an account inside a wallet.
+data LocalAccountReference
```
maksbotan: Why can't we reuse `AccountReference`?

gromakovsky: Because `AccountReference` contains `WalletReference` (`AccountRefByIndex` and `AccountRefByName` constructors). I want to ensure that all accounts are from the same wallet. If I had a list of `AccountReference`s, I wouldn't statically know that all of them have the same `WalletReference` (I could only write comments about it).

docs/usage-tui.md
gromakovsky: I think it's ok for TUI users :man_shrugging:

docs/usage-tui.md
gromakovsky: > * If no accounts are explicitly specified and an account from the
>   input wallet is selected, only its addresses will be used as inputs.

Isn't it the same? :thinking:

maksbotan: Okay, I read some other words in my head.

PR #200, opened 2018-06-20
[WIP] [AD-239] Fix and enable knit tests

gromakovsky:


knit/src/Knit/Core.hs
```diff
@@ -93,9 +93,21 @@ instance Elem components Core => ComponentTokenizer components Core where
...
+        -- parser for a single character:
+        -- `charLiteral <|> anyChar`
+        -- It didn't account for a special character combination `\&`, which
+        -- is like absence of a character.
+        -- New parser for a character returns `Maybe Char`, because for `\&`
+        -- there is no character.
+        let pChar :: Tokenizer (Maybe Char)
+            pChar =
+              (Just <$> P.charLiteral) <|>
+              (Nothing <$ P.string "\\&") <|>
```
int-index: Yes that's the fix I was talking about

PR #201, opened 2018-06-21
Hotfix universum bump

gromakovsky:
`putText` has different semantics: it doesn't print newline character anymore.
`putStrLn` does print newline character.


vrom911:
FYI, there is [`putTextLn`](https://hackage.haskell.org/package/universum-1.2.0/docs/Universum-Print.html#v:putTextLn) in `universum` for printing `Text` with newline.

gromakovsky:
Yes, I know, but I prefer using `putStrLn` whenever type of the argument is concrete, because this name is more standard and is probably more understandable to people who are not familiar with `universum` very well.

PR #204, opened 2018-06-24
[AD-222] Change arrow keys behavior in Tree

maksbotan:
Up and Down now scroll through all items, old behavior is preserved in
Ctrl-Up and Ctrl-Down.

---

There are two disputable decisions in this PR:

- I made `KeyUp` and `KeyDown` events fire only when no modifier is presed. Previously REPL, for example, would treat Up and Ctrl-Up the same way, now it will react only to Up. Is that OK?
- I changed `j` and `k` to work like new Up and Down, not like new Ctrl-Up and Ctrl-Down (the old behavior).

### DIFF
--- a/docs/usage-tui.md
+++ b/docs/usage-tui.md
@@ -48,9 +48,10 @@ The tree can be navigated in several ways. First of all, you can select any entr
 with mouse. You can also change the current selection with the keyboard. Up and Down keys will go to
 the previous and the next item respectively, including the `[ + Add wallet ]` button. Right key will
 go to the first child of the current item and Left key will go to the parent. Ctrl-Up and Ctrl-Down
-can be used to switch between items of the same type &mdash; e.g. go to the next or previous wallet,
-if a wallet is currently selected. Finally, `h`, `j`, `k`, `l` keys will work exactly as arrow keys
-(without Ctrl).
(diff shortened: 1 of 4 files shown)


goodsoft:
Will you manage to fix `←` key alternating between first wallet and `Add Wallet`, and `Ctrl+↑` not going up to `Add Wallet`?

maksbotan:
Fixed left key, please test (I did not wait for it to build). For Ctrl-Up on first wallet - I don't think it should behave this way, but would be trivial to fix if needed. @gromakovsky ?

goodsoft:
```
src/Ariadne/UI/Vty/Widget/Tree.hs:229:69: error:
    • Couldn't match expected type ‘[Word]’
                  with actual type ‘NonEmpty Word’
    • In the expression: p
      In the expression: if length p == 1 then p else init p
      In the second argument of ‘($)’, namely
        ‘\ p -> if length p == 1 then p else init p’
    |
229 |   TreeNavigationParent -> defaultPath $ \p -> if length p == 1 then p else NE.init p
| ^
```
Phale.

maksbotan:
So much for hie. *Sigh*

gromakovsky:
I tested locally and it seems to work as we want.

PR #207, opened 2018-06-26
[AD-219] Introduce BIP-44 helpers

Pastafarianist:
* Introduce `Word31`, a custom type similar to `Coin`, that has `maxBound = 2^31-1`. It lives in `Ariadne.Wallet.Cardano.Kernel.Word31`.
* Put all BIP-44-related conversion into a separate module (`Ariadne.Wallet.Cardano.Kernel.Bip44`) and only expose the necessary stuff outside.
* Modify the rest of the code to comply with the refactoring.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
+++ b/ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
@@ -1,0 +1,60 @@
+module Ariadne.Wallet.Cardano.Kernel.Word31
+    ( Word31
+    , Word31Exception (..)
+    , unsafeMkWord31
+    , word31ToWord32
+    ) where
+
+import Universum
+
+import Control.Monad.Except (MonadError(throwError))
+import Data.SafeCopy (base, deriveSafeCopySimple)
+import qualified Data.Text.Buildable (Buildable (..))
+import Formatting (bprint, int)
+
+{-------------------------------------------------------------------------------
+  Supporting types
+-------------------------------------------------------------------------------}
+
+newtype Word31 = Word31
+    { getWord31 :: Word32
+    } deriving (Show, Ord, Eq)
+
+newtype Word31Exception = Word31Overflow Word32
+    deriving (Show)
+
+instance Exception Word31Exception where
+    displayException (Word31Overflow n) =
+        "Word31: " <> show n <> " is too large"
+
+maxWord31 :: Word32
+maxWord31 = 0x7FFFFFFF -- 2^31 - 1
+
+instance Bounded Word31 where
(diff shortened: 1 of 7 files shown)


ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
```diff
@@ -0,0 +1,60 @@
...
+
+maxWord31 :: Word32
+maxWord31 = 0x7FFFFFFF -- 2^31 - 1
+
+instance Bounded Word31 where
+    minBound = Word31 0
+    maxBound = Word31 maxWord31
+
+-- | Makes a 'Word31' but is _|_ if that number exceeds 2^31 - 1
+unsafeMkWord31 :: Word32 -> Word31
```
hithroc: Why only have an unsafe version? Why not have `Word32 -> Maybe Word31` or something?

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
```diff
@@ -0,0 +1,60 @@
...
+unsafeMkWord31 n =
+    either (error . toText . displayException) (const word31) (checkWord31 word31)
+  where
+    word31 = Word31 n
+{-# INLINE unsafeMkWord31 #-}
+
+checkWord31 :: MonadError Word31Exception m => Word31 -> m ()
```
hithroc: This is a predicate. I don't understand really understand why do we need `MonadError Word31Exception m` construction for a predicate. 
And why do we need `Word31Exception` at all and why is it exported from the module? It is used only within the module in an internal function that is not exported. This function is only used in `unsafeMkWord31` which immediately converts this exception into `error` which makes it all look very redundant to me.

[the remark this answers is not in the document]
Pastafarianist: In this case I blindly copied the implementation of `Pos.Core.Common.Coin`. Ok, I will refactor this one.

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
Pastafarianist: It's not needed anywhere :man_shrugging:

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
Pastafarianist: Again, this is due to the fact that I blindly copied a part of `Pos.Core.Common.Coin`.

PR #208, opened 2018-06-26
[AD-235] Add InputSelectionPolicy argument to the `send` command

gromakovsky:


ariadne/src/Ariadne/Wallet/Knit.hs
```diff
@@ -4,28 +4,47 @@ import Universum
...
 
 import Knit
 
 -- Component type for Knit
 data Wallet
 
-data instance ComponentValue _ Wallet
+data instance
+     ComponentValue _
+       Wallet = ValueInputSelectionPolicy !InputSelectionPolicy
```
hithroc: This formatting is a little bit weird if you ask me.

gromakovsky: `hindent` :man_shrugging:

PR #211, opened 2018-06-28
[AD-135] Make BIP-32 functions more granular and reusable

gromakovsky:
1. Part of BIP-32 logic was moved from KeyStorage.hs to a new module,
   so that it logic can be easier used from other places.
2. Previously there was a function to derive a key by path and construct
   an address (all at once). I split it into two functions basically.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -49,8 +49,6 @@ import Ariadne.Config.Wallet (WalletConfig(..))
 import Ariadne.Wallet.Cardano.Kernel.Bip32
 import Ariadne.Wallet.Cardano.Kernel.Bip44
   (Bip44DerivationPath(..), encodeBip44DerivationPath)
-import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet
-  (HdAccountIx(..), HdAddressChain(..), HdAddressIx(..))
 import Ariadne.Wallet.Face
 
 data NoWalletSelection = NoWalletSelection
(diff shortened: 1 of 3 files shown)


ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -503,7 +504,7 @@ findFirstUnique lastIdx paths = head
   where
     pathsSet = V.foldr S.insert S.empty paths
 
--- ^ This function derives a 3-level address using account index, change index
+-- | This function derives a 3-level address using account index, change index
```
Pastafarianist: pofiksheno

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Bip32.hs
```diff
@@ -0,0 +1,42 @@
...
+    -> Maybe EncryptedSecretKey
+deriveHDSecretKeyByPath shouldCheck pp sk derPath =
+    foldM
+        (\sk' (check, idx) -> deriveHDSecretKey check pp sk' idx)
+        sk
+        -- check passphrase at most once
+        (zip (shouldCheck : repeat (ShouldCheckPassphrase False)) derPath)
+
+-- | Like 'makePubKeyHdwAddress', but also creates HDPassphrase internally.
+makePubKeyHdwAddress' ::
```
Pastafarianist: Can you make this name more descriptive? Maybe `makePubKeyHdwAddressCreatingPassphrase` (we're venturing into Java territory here, aren't we?)

ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -514,25 +515,18 @@ deriveBip44KeyPair ::
...
-    return (makePubKeyHdwAddress era hdPayload (encToPublic addrKey), addrKey)
+deriveBip44KeyPair era pp rootSK hdAccountIx hdAddressChain hdAddressIx =
+    toPair <$>
+    deriveHDSecretKeyByPath (ShouldCheckPassphrase True) pp rootSK derPath
   where
-    mAddrKey = foldM
-        (\sk' (check, idx) -> deriveHDSecretKey (ShouldCheckPassphrase check) pp sk' idx)
-        sk
-        (zip (True:repeat False) derPath) -- check passphrase only once
+    bip44DerPath = Bip44DerivationPath hdAccountIx hdAddressChain hdAddressIx
```
Pastafarianist: Maybe we should pack the 3 parameters of this function into one `Bip44DerivationPath`, so that you don't have to write this?

gromakovsky: Makes sense, will do.

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Bip32.hs
gromakovsky: I couldn't come up with a not-too-long name :/
`makePubKeyHdwAddressCreatingPassphrase` looks too long (I am the person who added `encodeBip44DerivationPathFromAccount` earlier today, I know :D)

`makePubKeyHdwAddressUsingPath` is shorter and is hopefully descriptive enough.

PR #214, opened 2018-07-01
[AD-182] Refactor widgets and event handling

goodsoft:
PR summary:
* [AD-211] Base widget:
  * existentially parametrized by custom widget state record
  * allows implementing just a subset of drawing/event handling methods
  * can contain arbitrary subwidgets
  * is automatically traversed when handling events
* [AD-212] Widget names are now hierarchical.
    This allows dispatching Brick events across the widget tree and keeping unique names even when we use the same widget in multiple places.
    Currently it's still one global enum, but in the future it should be made a union type, so that each widget can locally define its children names.
* [AD-210] Button, checkbox and edit widgets implemented. Features include:
  * Full integration with our keyboard scheme (not possible with Brick's included widgets)
  * No event dispatching boilerplate in parent widgets needed anymore
  * Multiline and password widgets
  * REPL is now implemented as custom edit widget to reduce code and logic duplication

These tasks have also been done as part of this PR:
* [AD-232] Multiline commands now correctly interact with REPL history
* [AD-231] REPL correctly handles empty commands
* [AD-213] Bracketed paste mode now works in all edit widgets, so it is reenabled

All existing widgets and `App` were refactored to be based on the new `Widget`.
In the process some minor changes to existing functionality were made:
* REPL output is now excluded from tab cycle and cannot be focused with `Esc-O`. Output can still be scrolled by `PgUp`/`PgDn` and `Shift/Ctrl+↑/↓` while focused in REPL input.
* `KeyCtrlUp/Down` were changed to `KeyModUp/Down`, so only plain `↑`/`↓` are now used for default up/down actions, and any modifier can be used for alternate action.

My implementation of `Widget` still has multiple minor issues, which can be solved later, when they start spoiling any fun:
* Monad stacks are weird :)
* Helper functions are somewhat arbitrary now, so there's still some manual lifting required sometimes.
* There's no direct parent→child communication except for lenses, as child becomes a "black box" right after instantiation
* child→parent events are all in one `WidgetEvent` enum, but should be split into widget-local types. This involves using `dependent-map` for registering listeners though.

Still, it removes a lot of event dispatching boilerplate and allows implementing reusable widgets, and remaining issues can be solved locally and incrementally when needed.


ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+
+addWidgetEventHandler :: MonadState (WidgetInfo s p) m => WidgetNamePart -> (WidgetEvent -> WidgetEventM s p ()) -> m ()
+addWidgetEventHandler namePart handler = do
+  widgetEventHandlersL %= Map.insert namePart handler
+
+setWidgetDraw :: MonadState (WidgetInfo s p) m => (s -> WidgetDrawM s p (B.Widget WidgetName)) -> m ()
+setWidgetDraw = assign widgetDrawL
+
+setWidgetDrawWithFocused :: MonadState (WidgetInfo s p) m => (Bool -> s -> WidgetDrawM s p (B.Widget WidgetName)) -> m ()
+setWidgetDrawWithFocused = assign widgetDrawWithFocusedL
```
maksbotan: Is there really no way to avoid this repetitive code?

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+  -> WidgetEventM s p a
+  -> StateT (Widget p) (StateT p (B.EventM WidgetName)) a
+withWidget widget@WidgetInfo{..} action = do
+  let
+    subaction = do
+      (res, widgetState') <- runStateT action widgetState
+      widgetStateL .= widgetState'
+      forM_ widgetEventQueue $ \(namePart, event) -> do
+        whenJust (Map.lookup namePart widgetEventHandlers) $ \handler -> do
+          use widgetStateL >>= execStateT (handler event) >>= assign widgetStateL
```
maksbotan: `widgetStateL %= execStateT (handler event)` ?

ui/vty/src/Ariadne/UI/Vty/App.hs
```diff
@@ -32,103 +32,189 @@ import Ariadne.UI.Vty.Widget.Status
...
+    drawHelpScreen = drawScreen [drawChild WidgetNameHelp]
+    drawAboutScreen = drawScreen [drawChild WidgetNameAbout]
+    drawLogsScreen = drawScreen [drawChild WidgetNameLogs]
+
+  return $ case appScreen of
+    AppScreenWallet -> drawWalletScreen
+    AppScreenHelp -> drawHelpScreen
+    AppScreenAbout -> drawAboutScreen
+    AppScreenLogs -> drawLogsScreen
+  {-let
```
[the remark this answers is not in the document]
goodsoft: Forgotten old code, oops

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+
+type WidgetName = [WidgetNamePart]
+
+data WidgetEventResult
+  = WidgetEventHandled
+  | WidgetEventNotHandled
+
+data WidgetInfo s p = WidgetInfo
+  { widgetName :: !WidgetName
+  , widgetState :: s
```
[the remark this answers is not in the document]
goodsoft: Yes, because it's `error` by default, will add a comment

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+  , widgetChildren :: !(Map WidgetNamePart (Widget (WidgetInfo s p)))
+  , widgetFocusList :: ![WidgetNamePart]
+  , widgetEventHandlers :: !(Map WidgetNamePart (WidgetEvent -> WidgetEventM s p ()))
+  , widgetEventQueue :: ![(WidgetNamePart, WidgetEvent)]
+  , widgetEventSend :: WidgetEvent -> WidgetEventM s p ()
```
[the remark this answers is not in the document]
goodsoft: This looks like oops.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
goodsoft: We could probably just expose lenses, but I'm not sure yet.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
goodsoft: Yeah, probably

ui/vty/src/Ariadne/UI/Vty/Widget.hs
goodsoft: I would rather leave it as-is, because this way we can easily change inner workings without rewriting all widget constructors.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
goodsoft: Nope, second argument of `%=` must be `s -> s`, but `execStateT (handler event)` is `s -> m s`

maksbotan: Okay, fair.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+    findInChild namePart = findClosestFocus [] <$> Map.lookup namePart widgetChildren
+findClosestFocus (np:nps) widget@(Widget WidgetInfo{..})
+  | Just child <- Map.lookup np widgetChildren = findClosestFocus nps child
+  | otherwise = findClosestFocus [] widget
+
+liftBrick :: B.EventM WidgetName a -> WidgetEventM s p a
+liftBrick = lift . lift . lift
+
+widgetParentGetter :: (p -> a) -> (WidgetInfo p q -> a)
+widgetParentGetter f = \WidgetInfo{ widgetState } -> f widgetState
```
maksbotan: nitpick: that's `view $ widgetStateL . to f`

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+
+widgetParentLens :: Lens' p a -> Lens' (WidgetInfo p q) a
+widgetParentLens l = lens getter setter
```
maksbotan: nitpick: that's `widgetStateL . l`

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
...
+data Widget p = forall s. Widget (WidgetInfo s p)
+
+instance B.Named (Widget p) WidgetName where
+  getName (Widget WidgetInfo{..}) = widgetName
+
+type WidgetInitM s p = State (WidgetInfo s p) ()
+
+type WidgetDrawM s p a = ReaderT (WidgetInfo s p) (Reader p) a
+
+type WidgetEventM s p a = StateT s (StateT (WidgetInfo s p) (StateT p (B.EventM WidgetName))) a
```
gromakovsky: So once again, why are nested `StateT`s and `ReaderT`s (above) necessary?

goodsoft: Because I want three things to be available:
* widget-local state `s`
* widget info `WidgetInfo s p`
* parent widget `p`
Moreover, I want lenses for `s` to just work.
The other things could be merged into a tuple or some other structure, but it would require more API methods to access them.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,394 @@
+module Ariadne.UI.Vty.Widget
```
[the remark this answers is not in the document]
goodsoft: Sure, will do that tomorrow.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
gromakovsky: Did you consider using classy lenses? https://michaelxavier.net/posts/2016-04-03-Enterprise-Haskell-Pattern-Lensed-Reader.html

goodsoft: I didn't, as I didn't stumble on this article.
Looks promising, but should probably be left for a next round of refactoring, which will be much less invasive and more mechanical.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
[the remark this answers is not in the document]
goodsoft: Added some Haddocks, please tell me if there's anything else non-obvious there.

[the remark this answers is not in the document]
gromakovsky: It's a good practice to also document each module (top-level Haddock comment) :wink:

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,464 @@
...
+  | WidgetNameWalletSendAddress
+  | WidgetNameWalletSendAmount
+  | WidgetNameWalletSendPass
+  | WidgetNameWalletSendButton
+
+  | WidgetNameAccount
+  deriving (Eq, Ord, Show)
+
+-- | Unique widget name, describing its location in widget tree
+type WidgetName = [WidgetNamePart]
```
gromakovsky: Probably would be better to make it `NonEmpty` (if I got it right that it must not be empty).

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,464 @@
...
+    widgetChildrenL %= Map.insert namePart (rename $ Widget child')
+  where
+    rename :: Widget p -> Widget p
+    rename (Widget widget@WidgetInfo{..}) =
+      Widget widget
+        { widgetName = namePart : widgetName
+        , widgetChildren = Map.map rename widgetChildren
+        }
+
+setWidgetFocusList :: MonadState (WidgetInfo s p) m => [WidgetNamePart] -> m ()
```
gromakovsky: When we get back to the topic of nested `StateT`s, I'd like to also consider using concrete monads here instead of `MonadState` constraints.

ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -0,0 +1,464 @@
...
+  , widgetChildren :: !(Map WidgetNamePart (Widget (WidgetInfo s p)))
+  , widgetFocusList :: ![WidgetNamePart]
+  , widgetEventHandlers :: !(Map WidgetNamePart (WidgetEvent -> WidgetEventM s p ()))
+  -- ^ Should be refactored into a dependent-map, for something like this:
+  -- @Map (e, WidgetNamePart) -> (e -> m ())@
+  , widgetEventQueue :: ![(WidgetNamePart, WidgetEvent)]
+  -- ^ List of events received from child widgets, is traversed after child event handlers complete
+  , widgetEventSend :: !(WidgetEvent -> WidgetEventM s p ())
+  -- ^ Send event to parent widget's queue
+  , widgetDraw :: !(s -> WidgetDrawM s p (B.Widget WidgetName))
```
maksbotan: (nitpick)
Why do you need to pass `s` as argument when `WidgetDrawM` is a reader over this struct, so one can define

```haskell
getState :: WidgetDrawM s p s
getState = view widgetStateL
```

PR #215, opened 2018-07-02
[AD-135] Support hasPass and hardcode assurance

Pastafarianist:


ariadne/src/Ariadne/Wallet/Backend/Util.hs
```diff
@@ -0,0 +1,22 @@
...
+import Ariadne.Cardano.Face (CardanoMode)
+import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet (HasSpendingPassword(..))
+import Ariadne.Wallet.Cardano.Kernel.DB.InDb (InDb(..))
+
+mkHasPass
+    :: (CardanoMode ~> IO)
+    -> Bool
+    -> IO HasSpendingPassword
+mkHasPass runCardanoMode noSpendingPassword =
+  if noSpendingPassword then
```
gromakovsky: https://github.com/serokell/ariadne/blob/master/code-style.md#if-then-else-clauses

PR #217, opened 2018-07-02
[AD-131] Make history file path configurable

mitchellvitez:
Decided to use general config for this rather than UI-specific (as mentioned in YT)


ariadne/src/Ariadne/Config/History.hs
```diff
@@ -0,0 +1,43 @@
...
+parseFieldHistory = parseField historyFieldModifier
+
+historyFieldModifier :: D.Text -> D.Text
+historyFieldModifier = f
+  where
+    f "hcPath" = "path"
+    f x = x
+
+data HistoryConfig = HistoryConfig
+  { hcPath :: Text
```
gromakovsky: Why is it `Text` and not `FilePath`?

PR #219, opened 2018-07-03
[AD-254] Allow multiple outputs for transactions

goodsoft:


ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -163,59 +187,59 @@ handleWalletWidgetEvent langFace@UiLangFace{..} = \case
...
+  inputs <- fmap (\WalletSendOutput{..} -> (walletSendAddress, walletSendAmount)) <$> uses walletSendOutputsL Map.elems
   passphrase <- use walletSendPassL
   use walletSendResultL >>= \case
     SendResultWaiting _ -> return ()
-    _ -> liftIO (langPutUiCommand $ UiSend address amount passphrase) >>=
+    _ -> liftIO (langPutUiCommand $ UiSend inputs passphrase) >>=
       assign walletSendResultL . either SendResultError SendResultWaiting
+
+unsafeFromJust :: Lens' (Maybe a) a
+unsafeFromJust = lens fromJust setJust
```
maksbotan: I regret spending time on that, since I couldn't improve this code.
One possible way is to use `Traversal'` instead of `Lens'`, as it allows for zero or more values, then you could use `. _Just .` unstead of `unsafeFromJust`. But that probably isn't worth it.

ui/vty/app/Glue.hs
```diff
@@ -102,16 +102,17 @@ knitFaceToUI UiFace{..} KnitFace{..} =
       UiBalance ->
         Right $ Knit.ExprProcCall
           (Knit.ProcCall Knit.balanceCommandName [])
-      UiSend address amount passphrase -> do
-        argAddress <- decodeTextAddress address
-        argCoin <- readEither amount
+      UiSend inputs passphrase -> do
```
maksbotan: Parameter says `inputs`, but you are constructing `tx-out`. Huh?

goodsoft: It's nice when someone actually reads your code

ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -163,59 +187,59 @@ handleWalletWidgetEvent langFace@UiLangFace{..} = \case
...
+      initEditWidget $ widgetParentLens $ walletSendOutputsL . at idx . unsafeFromJust . walletSendAmountL
+    addWidgetChild (WidgetNameWalletSendRemove idx) $
+      initButtonWidget "-"
+    addWidgetEventHandler (WidgetNameWalletSendRemove idx) $ \case
+      WidgetEventButtonPressed -> removeOutput idx
+      _ -> return ()
+
+removeOutput :: Int -> WidgetEventM WalletWidgetState p ()
+removeOutput idx = do
+  walletSendOutputsL %= Map.delete idx
```
maksbotan: Maybe disallow deleting the only output?

ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -152,26 +184,78 @@ handleWalletWidgetEvent = \case
...
+    addWidgetChild (WidgetNameWalletSendAmount idx) $
+      initEditWidget $ widgetParentLens $ walletSendOutputsL . at idx . unsafeFromJust . walletSendAmountL
+    addWidgetChild (WidgetNameWalletSendRemove idx) $
+      initButtonWidget "-"
+    addWidgetEventHandler (WidgetNameWalletSendRemove idx) $ \case
+      WidgetEventButtonPressed -> removeOutput idx
+      _ -> return ()
+
+removeOutput :: Int -> WidgetEventM WalletWidgetState p ()
+removeOutput idx = do
```
maksbotan: what happens to the widgets with their handlers, etc? do we need to remove them as well?

goodsoft: We should implement proper children removal at some point, but currently they don't do any harm, as they don't get any focus, so can't handle any events.

PR #221, opened 2018-07-04
[OPS-322] fail on hlint warnings

yorickvP:


gromakovsky:
Default hlint rules are usually not suitable out-of-box. We need to come up with ariadne-specific rules first.

Pastafarianist:
We'll have to review these hints later. For now, do I understand correctly that `hlint` does not currently run on `ariadne-vty`? If so, can you enable it?

maksbotan:
Would be great if you somehow automatically annotaded PR lines, in addition to just providing the log. In the current form it's not very comfortable to use.

yorickvP:
I considered exposing the html report, but it would not be nearly as good
as people running hlint themselves. The main purpose of doing it in the CI
is to make it fail when there are errors.
GitHub supports some annotation thing nowadays, I wouldn't count on devtime
being spent on using that.

Pastafarianist:
Blocked until https://issues.serokell.io/issue/AD-285 is resolved.

mkaito:
When merged, update https://issues.serokell.io/issue/OPS-322

Pastafarianist:
Updated the issue. Also commenting here to ensure that this does not slip through the cracks.

gromakovsky:
AFAIU it's done and the issue can be closed.

PR #194, opened 2018-06-17
[AD-135] Switch to the new backend

Dolgakov:
- [x] Finish some stuff in KeyStorage.hs and Glue.hs
- [x] Insert imports, make it compile.
- [x] Write restoreFromSecretKey.

PR #222, opened 2018-07-04
[AD-135] Simple improvements necessary to use new backend

gromakovsky:
These changes are from #194. They're quite standalone, simple and self-contained, so I decided to move it into a separate PR to make #194 a bit smaller and easier to review (it's very big currently).

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
+++ b/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
@@ -17,6 +17,7 @@ module Ariadne.Wallet.Cardano.Kernel.DB.AcidState (
     -- ** Updates on HD wallets
     -- *** CREATE
   , CreateHdWallet(..)
+  , CreateHdRoot(..)
   , CreateHdAddress(..)
     -- *** UPDATE
   , UpdateHdRootAssurance
--- a/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
+++ b/ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
@@ -85,23 +85,29 @@ import Ariadne.Wallet.Cardano.Kernel.Word31 (Word31)
   Supporting types
 -------------------------------------------------------------------------------}
 
--- | Wallet name
-newtype WalletName = WalletName Text
+-- | Name of a wallet.
+newtype WalletName = WalletName
+    { unWalletName :: Text
+    } deriving (Show, Eq, Ord, IsString, ToString)
 
 -- | Account name
-newtype AccountName = AccountName Text
+newtype AccountName = AccountName
+    { unAccountName :: Text
+    } deriving (Show, Eq, Ord, IsString, ToString)
 
 -- | Account index
-newtype HdAccountIx = HdAccountIx Word31
-  deriving (Eq, Ord, Show)
+newtype HdAccountIx = HdAccountIx
+    { unHdAccountIx :: Word31
+    } deriving (Eq, Show, Ord)
 
 -- | Whether the chain is an external or an internal one
 data HdAddressChain = HdChainExternal | HdChainInternal
   deriving (Eq, Ord, Show)
 
 -- | Address index
-newtype HdAddressIx = HdAddressIx Word31
-  deriving (Eq, Ord, Show)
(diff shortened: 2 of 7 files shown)


ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -266,6 +268,11 @@ createHdAddress :: HdAddress -> Update DB (Either HD.CreateHdAddressError ())
 createHdAddress hdAddress = runUpdate' . zoom dbHdWallets $
     HD.createHdAddress hdAddress
 
+createHdAccount :: HdAccount
```
Pastafarianist: As far as I remember, this does not have to exist or be exposed outside. The replacement is a different function also called `createHdAccount` with the following signature:

```
createHdAccount :: HdAccountId
                -> PrefilteredUtxo
                -> Maybe AccountName
                -> Update DB (Either HD.CreateHdAccountError ())
```

Can you explain where you need this?

ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
```diff
@@ -132,23 +139,27 @@ deriveSafeCopySimple 1 'base ''HasSpendingPassword
 -------------------------------------------------------------------------------}
 
 -- | HD wallet root ID
-data HdRootId = HdRootId (InDb (Core.AddressHash Core.PublicKey))
-  deriving (Eq, Ord)
+data HdRootId = HdRootId
```
Pastafarianist: Maybe replace this with a `newtype` and use `GeneralizedNewtypeDeriving` for `Show`?

ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet/Create.hs
```diff
@@ -56,6 +58,25 @@ data CreateHdAddressError =
 
     -- | Address already used
   | CreateHdAddressExists HdAddressId
+  deriving (Eq, Show)
+
+instance Exception CreateHdRootError where
+  displayException (CreateHdRootExists rootId) =
+    "The wallet " ++ show rootId ++ " is already exist."
```
Pastafarianist: > is exist

same thing a few lines below

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
```diff
@@ -26,6 +26,10 @@ instance Exception Word31Exception where
     displayException (Word31Overflow n) =
         "Word31: " <> show n <> " is too large"
 
+instance Enum Word31 where
+    toEnum = unsafeMkWord31 . fromIntegral
```
Pastafarianist: Are you sure `unsafeMkWord31` should be used here? Where is this instance needed anyway?

gromakovsky: I started checking it and ended up doing `stack clean --full` :unamused:

gromakovsky: Anyway, `unsafeMkWord31` is ok, because `toEnum` is partial anyway. It's better to add a comment about it, though.

ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
Pastafarianist: Yes, but it says `hdAccId`, not `hdAcc`, so it refers to the function with the signature I provided.

ariadne/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
Pastafarianist: :ok_hand:

ariadne/src/Ariadne/Wallet/Cardano/Kernel/Word31.hs
gromakovsky:
```
/home/gromak/Work/ariadne/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs:298:9: error:
    • No instance for (Enum Word31)
        arising from a use of ‘findFirstUnique’

```

PR #224, opened 2018-07-05
[AD-252] List accounts in Wallet widget, allow selecting them as transaction inputs

goodsoft:


maksbotan:
This should be rebased after AD-254 is merged.

PR #228, opened 2018-07-06
[AD-257] Allow renaming wallet in Wallet widget

goodsoft:


ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -177,7 +193,14 @@ drawWalletWidget focus WalletWidgetState{..} = do
     B.padAll 1 $
     B.vBox $
     padBottom <$>
-      [ label "Wallet name:" B.<+> B.txt walletName
+      [ label "Wallet name:"
+          B.<+> drawChild WidgetNameWalletName
+          B.<+> padLeft (drawChild WidgetNameWalletRenameButton)
+      , case walletRenameResult of
+          RenameResultNone -> B.emptyWidget
```
maksbotan: Maybe we will consider abstracting something for widgets that perform backend commands in the future...

ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -291,6 +323,15 @@ updateFocusList = do
       , WidgetNameWalletSendRemove idx
       ]
 
+performRename :: WidgetEventM WalletWidgetState p ()
+performRename = do
```
maksbotan: And this repeating stuff starts to look too boilerplate as well, but I don't think it should be generalized at the moment.

ui/vty/src/Ariadne/UI/Vty/Widget/Wallet.hs
goodsoft: Yeah, I thought about that too.
That's also the reason I try to make this code as similar looking as possible, so that it can be easily and mechanically refactored in the future.

PR #232, opened 2018-07-09
[AD-264] Add balance to tree in UiAdapter

Dolgakov:
`UiWalletData` tree now have balances in all levels.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/UiAdapter.hs
+++ b/ariadne/src/Ariadne/Wallet/UiAdapter.hs
@@ -2,9 +2,21 @@ module Ariadne.Wallet.UiAdapter
   ( UiWalletData(..)
   , UiAccountData(..)
   , UiWalletSelection(..)
+  , UiAddressData(..)
   -- * Lens
+  , uwdName
+  , uwdId
   , uwdAccounts
+  , uwdBalance
+
+  , uadName
+  , uadAccountIdx
   , uadAddresses
+  , uadBalance
+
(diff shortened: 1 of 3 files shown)


goodsoft:
Works for me, thanks.
Not putting a seal of approval, as I'm not familiar enough with balance calculation logic.

Pastafarianist:
The logic is that `hd{Root, Account, Address}Balance` should return what you expect, with the exception of `hdRootBalance` which returns `Integer` instead of `Coin`. We'll probably fix that later.

ariadne/src/Ariadne/Wallet/UiAdapter.hs
```diff
@@ -21,26 +32,36 @@ import Ariadne.Wallet.Cardano.Kernel.DB.InDb
...
+  , _uwdBalance  :: !Coin
   } deriving (Show, Generic)
 
 data UiAccountData = UiAccountData
   { _uadName      :: !Text
   , _uadPath      :: !Word32
   -- Because of ChainType layer Now it should be ((HdAddressChain, Word32), Address) I guess, but
   -- AFAIU we don't want a new layer, so addresses of both types wiil be in one list -- External first.
-  , _uadAddresses :: !(Vector (Word32, Address))
+  , _uadAddresses :: !(Vector (Word32, UiAddressData))
```
gromakovsky: Is there a reason not to put `Word32` into `UiAddressData` to make types simpler?

PR #233, opened 2018-07-09
[AD-263] Show balances in list of addresses in account

goodsoft:
Depends on #232, so it should be merged first


ui/vty/app/Glue.hs
```diff
@@ -265,7 +257,7 @@ walletEventToUI = \case
       UiWalletUpdate
         (uiWalletDatasToTree (toUiWalletDatas db))
         (uiWalletSelectionToTreeSelection . (toUiWalletSelection db) <$> sel)
-        ((walletSelectionToPane (toUiWalletDatas db)) . (toUiWalletSelection db) <$> sel)
+        ((walletSelectionToInfo (toUiWalletDatas db)) . (toUiWalletSelection db) <$> sel)
```
maksbotan: LISP

PR #236, opened 2018-07-10
[AD-179] applyBlock for address checkpoints

Pastafarianist:
**YT issue:** https://issues.serokell.io/issue/AD-179

**Checklist:**

- [x] Adressed HLint warnings and hints _(irrelevant for imported code)_

**Description:**

There are multiple essential changes in this PR:

1. `PrefilteredBlock` is now more structured. Previously there used to be just `Set TxIn` and `Utxo`; now the same things are stored in maps indexed by `AddrWithId`.
2. The stuff in `AcidState.hs` now passes around a function that narrows down a `PrefilteredBlock` or `PrefilteredUtxo` to just one address. It's named `narrowP` in `createPrefiltered`.
3. All of the former allows to implement `applyBlock` for address checkpoints. It reuses the exact same code as account checkpoints, but takes as input only the data that is relevant to the address that is being processed ("data" meaning `TxIn`s referring to spendings from this address and new utxo belonging to this address).

Also, some minor cosmetic changes are introduced (like indentation, parentheses, etc). I hope at some point we will adopt `brittany` or perhaps even `hindent`, because the code we imported from IOHK is very inconsistent in terms of style.

Still TODO:

- [x] Implement `Buildable` instance for `PrefilteredBlock`. I ran into some issues due to the fact that the code uses `type` instead of `newtype`, which makes it difficult to avoid orphan instances. This is not urgent and perhaps the commented-out instance can even be deleted for now, as currently this instance is never used.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Cardano/Kernel/PrefilterTx.hs
+++ b/ariadne/src/Ariadne/Wallet/Cardano/Kernel/PrefilterTx.hs
@@ -4,6 +4,8 @@ module Ariadne.Wallet.Cardano.Kernel.PrefilterTx
        ( PrefilteredBlock(..)
        , PrefilteredUtxo
        , AddrWithId
+       , pfbInputs
+       , pfbOutputs
        , prefilterBlock
        , prefilterUtxo
        ) where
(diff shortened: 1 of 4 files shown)


ariadne/src/Ariadne/Wallet/Cardano/Kernel/PrefilterTx.hs
```diff
@@ -195,12 +205,23 @@ prefilter (wid,hdPass) selectAddr rtxs
...
-instance Buildable PrefilteredBlock where
-  build PrefilteredBlock{..} = bprint
-    ( "PrefilteredBlock "
-    % "{ inputs:  " % listJson
-    % ", outputs: " % mapJson
-    % "}"
-    )
-    (Set.toList pfbInputs)
-    pfbOutputs
+-- TODO: how to avoid introducing orphans here?
```
gromakovsky: It's not necessary to define `instance Buildable AddrWithId` in order to define `Buildable Prefiltered{Block,Utxo}`.

gromakovsky:
Shouldn't the second commit be tagged with AD-255?

PR #239, opened 2018-07-10
[AD-233] Consider accounts argument in 'sendTx'

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-233

**Checklist:**

  - [x] [README](README.md)

**Description:**
This PR finishes AD-233. Earlier I added an argument with list of accounts to be used as inputs, but this argument was ignored.

Didn't test yet, will do. **Upd**: tested, seems to work.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -5,6 +5,7 @@ module Ariadne.Wallet.Backend.KeyStorage
          -- * Commands/other functions
          refreshState
        , resolveWalletRef
+       , resolveWalletRefThenRead
        , newAddress
        , newAccount
        , newWallet
@@ -171,6 +172,22 @@ resolveWalletRef walletSelRef walRef db = case walRef of
 
     walletList :: [HdRoot]
     walletList = toList (readAllHdRoots hdWallets)
+
+-- | Like 'resolveWalletRef', but also reads some data from 'DB'
+-- corresponding to the resolved root ID, assuming that the wallet
+-- with this ID definitely exists.
+resolveWalletRefThenRead
+  :: IORef (Maybe WalletSelection)
+  -> WalletReference
+  -> DB
+  -> (HdRootId -> HdQueryErr UnknownHdRoot a)
+  -> IO (HdRootId, a)
+resolveWalletRefThenRead walletSelRef walRef db q = do
+    rootId <- resolveWalletRef walletSelRef walRef db
+    (rootId,) <$> case q rootId (db ^. dbHdWallets) of
+        -- This function's assumption is that it must not happen.
+        Left err -> error $ "resolveWalletRefThenRead: " <> pretty err
+        Right res -> return res
 
 resolveAccountRef
   :: IORef (Maybe WalletSelection)
--- a/ariadne/src/Ariadne/Wallet/Backend/Tx.hs
+++ b/ariadne/src/Ariadne/Wallet/Backend/Tx.hs
@@ -11,10 +11,10 @@ import qualified Data.List.NonEmpty as NE
 import qualified Data.Text.Buildable
 
 import Control.Exception (Exception(displayException))
-import Control.Lens (at)
+import Control.Lens (at, ix)
 import Data.Acid (AcidState, query)
 import Data.Map (findWithDefault)
-import Formatting (bprint, build, formatToString, (%))
+import Formatting (bprint, build, formatToString, int, (%))
 import Text.PrettyPrint.ANSI.Leijen (Doc, list, softline, string)
 
 import Pos.Client.KeyStorage (getSecretDefault)
@@ -37,13 +37,14 @@ import Ariadne.Wallet.Cardano.Kernel.DB.AcidState
 import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet
 import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Read
 import Ariadne.Wallet.Cardano.Kernel.DB.InDb
-import Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet (IxSet)
+import Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet (IxSet, (@+))
 import Ariadne.Wallet.Face
 import IiExtras ((:~>)(..))
 
 data SendTxException
     = SendTxNoAddresses !HdRootId
     | SendTxIncorrectPassPhrase
+    | SendTxAccountIndexOutOfRange !HdRootId !Word
     deriving (Show)
 
 instance Buildable SendTxException where
@@ -52,6 +53,8 @@ instance Buildable SendTxException where
             bprint ("Wallet #"%build%" does not have any addresses") walletIdx
         SendTxIncorrectPassPhrase ->
             "Incorrect passphrase"
+        SendTxAccountIndexOutOfRange rootId idx ->
+            bprint ("Account #"%int%" doesn't exist in "%build) idx rootId
 
 instance Exception SendTxException where
     displayException = toString . prettyL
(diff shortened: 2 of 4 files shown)


ariadne/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -74,22 +77,55 @@ sendTx ::
...
+        case nonEmpty accRefs of
+            Nothing -> do
+                readIORef walletSelRef >>= \case
+                    Just (WSAccount selectedAccId)
+                        | selectedAccId ^. hdAccountIdParent == rootId ->
+                            return (Just $ one selectedAccId)
+                    _ -> return Nothing
+            Just accRefsNE -> Just <$> mapM refToId accRefsNE
+      where
+        -- Here we use knowledge of the order in UI, which is not very good.
```
Pastafarianist: I'm not even sure how the order of `IxSet` corresponds to the UI order. Could you explain that part?

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -74,22 +77,55 @@ sendTx ::
...
+    resolveAccountRefs ::
+           HdRootId
+        -> IxSet HdAccount
+        -> IO (Maybe (NonEmpty HdAccountId))
+    resolveAccountRefs rootId accountsSet =
+        case nonEmpty accRefs of
+            Nothing -> do
+                readIORef walletSelRef >>= \case
+                    Just (WSAccount selectedAccId)
+                        | selectedAccId ^. hdAccountIdParent == rootId ->
```
Pastafarianist: As far as I see, this `rootId` comes from `resolveWalletRef`, but here you are checking that `walletSelRef` contains a `HdAccountId` referring to the same `rootId` that has already been returned by `resolveWalletRef`. I haven't read what `resolveWalletRef` does in detail or where `walletSelRef` comes from, but are you sure that this extra check is needed here?

ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -172,6 +173,22 @@ resolveWalletRef walletSelRef walRef db = case walRef of
     walletList :: [HdRoot]
     walletList = toList (readAllHdRoots hdWallets)
 
+-- | Like 'resolveWalletRef', but also reads some data from 'DB'
+-- corresponding to the resolved root ID, assuming that the wallet
+-- with this ID definitely exists.
+resolveWalletRefThenRead
```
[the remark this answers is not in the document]
gromakovsky: No, this function is very useful. When we resolve wallet ref, we only get a `HdRootId`, but we most certainly want to do something with this id, probably read something from DB corresponding to this id.
Currently most of the code relies on the fact that `resolveWalletRef` returns id of a wallet that definitely exists in DB (where it's relevant). Which is more or less true, but can be wrong if concurrent modifications occur. AFAIU currently they can't occur and we also have AD-274. Anyway, if such inconsistency occurs we treat it as a programmer mistake.
Sadly, it's not easy to represent this information at the type level. This function does it to some extent, so that its caller doesn't need to handle a programmer mistake case.
Now, regarding usages. Yes, currently it's only used in `Tx.hs`, but that's merely because this PR is about `Tx.hs`, so I didn't bother to find other places where `error` calls can be eliminated using this function. In fact, I wanted to created a separate issue to eliminate some `error`s introduced by AD-135, but apparently I haven't done it yet.
Quick search showed me the following:
1. First usage of `resolveWalletRef` is in `resolveAccountRef` and there we read accounts for this wallet. If it fails, we throw an exception in `IO`. Which is more or less OK, but according to CSL guidelines which we don't explicitly follow (yet), but which are still good, programmer mistakes should be handled using `error`, not `IO` exceptions.
2. Second usage is in `newAccount`. There we have this:
```
        -- TODO: make it global
    -- The same function as in `select`. It is unsafe -> not global.
    -- It is assumed that hdRootId is a valid one.
    -- It is, because resolveWalletRef always returns a valid HdRootId.
    getAccounts :: HdWallets -> HdRootId -> IxSet HdAccount
    getAccounts wallets rootId = fromRight
      (error "Bug: UnknownHdRoot")
      (readAccountsByRootId rootId wallets)
```
My function allows to avoid this `error` call and deprecates this comment.
3. Next usage is in `select`. There we have the same `getAccounts`.

I didn't search further, because I suppose that's convincing enough already.
During more profound refactoring we should consider adding more such helpers (e. g. for `resolveAccountRef`), but I didn't think about it yet.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
gromakovsky: We have implicit contract that UI indices should be treated as indices in `toList @(IxSet HdSomething)`. I find it fragile and want to refactor it. I think that backend shouldn't know about UI indices at all and all commands should take stable identifiers. So `LocalAccountReference` should have a constructor with `HdAccountIx`, not `Word`. Then we'll be able to do lookup using primary key, not using `toList` and `ix`.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
gromakovsky: See the documentation of this function.
```
If list of accounts is not empty, only those accounts
may be used as inputs.  If this list is empty and an account from
the input wallet is selected, this account will be used as input.
Otherwise inputs will be selected from all accounts in the wallet.
```

This case is when an account from the input wallet is selected.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -74,22 +77,55 @@ sendTx ::
...
-        let
-            walletAccounts :: IxSet HdAccount
-            walletAccounts = fromRight
-                -- Resolved walletRootId always exists.
-                (error "SendTx: Unknown walletRootId")
-                (readAccountsByRootId walletRootId wallets)
-
+    -- Returns list of accounts which can be used.
+    -- 'Nothing' means all accounts can be used.
+    resolveAccountRefs ::
```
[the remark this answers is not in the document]
gromakovsky: I don't want to move it inside, because then I'll have to return too much stuff (`walletRootId` and `walletAccounts`) are also needed outside. I can rename this function to e. g. `getSuitableAccounts`.

ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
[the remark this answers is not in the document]
Pastafarianist: This sounds quite convincing, thank you.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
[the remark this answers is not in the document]
Pastafarianist: `getSuitableAccounts` sounds good to me.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
Pastafarianist: Ok, LGTM.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
gromakovsky: Hm, I thought it would be fixed by AD-168, but in fact AD-168 can fix it only for `select`, not for commands like `new-account` which use `tyWalletRef`.
I'll create a separate issue for this.

gromakovsky: https://issues.serokell.io/issue/AD-294

ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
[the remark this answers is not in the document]
gromakovsky: https://issues.serokell.io/issue/AD-302

PR #241, opened 2018-07-11
[AD-234] Generate a new address as change

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-234

**Checklist:**

  - [x] [README](README.md)

**Description:**

Previously we were reusing one of our addresses as change address,
which is considered bad practice.
Proper approach is to generate a new change address for each transaction.
This logic is handled by the `MonadAddresses` class.
For simplicity an `IO` action which generates a new address is used
as `AddrData`. This action is `walletNewAddress` from `WalletFace` with
proper arguments. `walletNewAddress` was modified to return the address
it generates.

**This PR targets another feature branch!** Should be rebased after #239 is merged.

### DIFF
--- a/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -247,7 +247,7 @@ newAddress ::
     -> AccountReference
     -> HdAddressChain
     -> PassPhrase
-    -> IO ()
+    -> IO Address
 newAddress acidDb WalletFace {..} walletSelRef runCardanoMode accRef chain pp = do
   walletDb <- query acidDb Snapshot
   accountId <- resolveAccountRef walletSelRef accRef walletDb
@@ -289,7 +289,7 @@ newAddress acidDb WalletFace {..} walletSelRef runCardanoMode accRef chain pp = 
       , _hdAddressCheckpoints = one emptyAddrCheckpoint
       }
   throwLeftIO $ update acidDb (CreateHdAddress hdAddress)
-  walletRefreshState
+  addr <$ walletRefreshState
     where
       -- Using the sequential indexation as in accounts.
       -- TODO:
--- a/ariadne/src/Ariadne/Wallet/Backend/Tx.hs
+++ b/ariadne/src/Ariadne/Wallet/Backend/Tx.hs
@@ -7,7 +7,6 @@ module Ariadne.Wallet.Backend.Tx
 import Universum hiding (list)
 
 import qualified Data.HashMap.Strict as HM
-import qualified Data.List.NonEmpty as NE
 import qualified Data.Text.Buildable
 
 import Control.Exception (Exception(displayException))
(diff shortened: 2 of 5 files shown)


ariadne/src/Ariadne/Wallet/Knit.hs
```diff
@@ -127,8 +127,7 @@ instance (Elem components Wallet, Elem components Core, Elem components Cardano)
             passphrase <- getPassPhraseArg
             pure (accountRef, chain, passphrase)
         , cpRepr = \(accountRef, chain, passphrase) -> CommandAction $ \WalletFace{..} -> do
-            walletNewAddress accountRef chain passphrase
-            return $ toValue ValueUnit
+            toValue . ValueAddress <$> walletNewAddress accountRef chain passphrase
```
Pastafarianist: I'm not sure how this change is relevant to the goal of the PR, but I guess it's a good change.

ariadne/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -289,7 +289,7 @@ newAddress acidDb WalletFace {..} walletSelRef runCardanoMode accRef chain pp =
       , _hdAddressCheckpoints = one emptyAddrCheckpoint
       }
   throwLeftIO $ update acidDb (CreateHdAddress hdAddress)
-  walletRefreshState
+  addr <$ walletRefreshState
```
Pastafarianist: Minor: personally, I believe using `<$` in non-`Functor` context is redundant because it requires remembering an extra operator whose equivalent is trivial:

```
walletRefreshState
pure addr
```

and arguably more readable.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -142,20 +144,27 @@ sendTx acidDb WalletFace {..} CardanoFace {..} walletSelRef printAction pp walle
...
                 (SendTxNoAddresses walletRootId)
                 (nonEmpty $ HM.keys signersMap)
-        let ourAddress = NE.head ourAddresses
+        -- We pick one of the accounts to generate change address in it.
+        ourAccount <-
+            maybeThrow
+                (SendTxNoAccounts walletRootId)
+                (toList accountsToUse ^? ix 0)
+        let ourAccountId = ourAccount ^. hdAccountId
+        let newChangeAddress = walletNewAddress
```
Pastafarianist: Minor: redundant `let`, can be merged into the previous one.

ariadne/src/Ariadne/Wallet/Knit.hs
gromakovsky: Not directly relevant indeed, but since I had to return `Address` from `walletNewAddress`, I decided that returning `ValueAddress` would be more reasonable than ignoring the result of `walletNewAddress` (this place had to be changed anyway, at least to avoid a warning).

goodsoft: Oh cool, now we can generate _and_ copy address to clipboard by one button press, as I intended.

ariadne/src/Ariadne/Wallet/Backend/Tx.hs
gromakovsky: Some people say that such `let`s are redundant, while others insist on using explicit `let` for each binding. Life is hard.
:man_shrugging: 
Btw, [our style guide says](https://github.com/serokell/ariadne/blob/master/code-style.md#let-expressions):

> Put let before each variable inside a do block

I don't like it and want to make a PR removing this rule.

PR #244, opened 2018-07-15
[AD-205] Improve default configuration approach

iokasimov:
Idea is simple - provide type `data Presence a = There a | File (Maybe FilePath)` to not break configuration files and leave an opportunity to define default values.

In tests I left File constructor (cause it should take some time to write generators for tests for There values).
All default instances are based on contents of config files.


Ariadne/src/Ariadne/Config/Presence.hs
```diff
@@ -0,0 +1,34 @@
+module Ariadne.Config.Presence
```
gromakovsky: This file is `Ariadne/src/Ariadne/Config/Presence.hs`, but should be `ariadne/src/Ariadne/Config/Presence.hs`

iokasimov: Sorry, I renamed the folder when I met a warning from `cabal`.

ariadne/src/Ariadne/Config/Presence.hs
```diff
@@ -0,0 +1,34 @@
...
+import Universum
+
+import Control.Lens (Prism', prism')
+
+data Presence a
+  = There a
+  | File (Maybe FilePath)
+  deriving (Eq, Show)
+
+instance Functor Presence where
```
gromakovsky: Can be derived using `DeriveFunctor`.

ariadne/src/Ariadne/Config/CLI.hs
```diff
@@ -513,9 +519,9 @@ cliConfigurationOptionsParser = do
     ]
   return CLI_ConfigurationOptions{..}
 
-cliCommonArgsParser :: Opt.Parser CLI_CommonArgs
+cliCommonArgsParser :: Opt.Parser (CLI_CommonArgs conf)
 cliCommonArgsParser = do
-  cli_logConfig <- optional $ strOption $ mconcat
+  cli_logConfig <- (<$>) File . optional . strOption . mconcat $
```
gromakovsky: It's a bit strange to see such usage of `(<$>)`, because there is `fmap` (same for any other operator which has a well known alias function), but ok.

ariadne/src/Ariadne/Config/Presence.hs
iokasimov: What for embed a file into memory if we have can just use default configuration? `There` is for this needs.

gromakovsky: How is it related to my suggestion to use `DeriveFunctor` instead of writing this instance manually? :thinking:

iokasimov: I like to write instances myself, it's one minute, but I understand what happens. The question about `There` is not related to this, but I don't know where I can answer on your question above. The goal of the task was to avoid using filepaths for default configurations, instead, we can define our default configuration values. But it can be used unless Cardano configuration types will be ripped out.

gromakovsky: You can write comments directly to this PR quoting the parts you are answering to.
I've written about the rest in this thread: https://serokell.slack.com/archives/C9B1DAMDJ/p1531476663000117

ariadne/src/Ariadne/Config/Cardano.hs
```diff
@@ -1,29 +1,333 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
gromakovsky: 1. Orphan `Eq` instances should be moved to `Ariadne.Cardano.Orphans`.
2. Orphan `Default` instances are not needed. `def :: X` can be turned into `defX :: X`. This type class doesn't give much benefit in this case. Moreover, it may cause some trouble if someone else defines `instance Default X` somewhere.

ariadne/src/Ariadne/Config/Cardano.hs
```diff
@@ -1,29 +1,333 @@
...
+    def = NodeConfiguration
+        { ccNetworkConnectionTimeout = 15000
+        , ccConversationEstablishTimeout = 30000
+        , ccBlockRetrievalQueueSize = 100
+        , ccPendingTxResubmissionPeriod = 7
+        , ccWalletProductionApi = False
+        , ccWalletTxCreationDisabled = False
+        }
+
+instance Default Configuration where
```
gromakovsky: No, this is wrong, this configuration won't work with mainnet. Mainnet configuration is very big, putting it into source code won't work. The idea is to embed configuration file (using `file-embed` probably) and parse it.

ariadne/src/Ariadne/Config/Cardano.hs
```diff
@@ -1,29 +1,333 @@
...
+    def = LoggerTree
+        { _ltSubloggers = mempty
+        , _ltFiles =
+            [ HandlerWrap "logs/node.pub" Nothing
+            , HandlerWrap "logs/node" Nothing
+            ]
+        , _ltSeverity = Just debugPlus
+        }
+
+instance Default LoggerConfig where
```
gromakovsky: Since you need to embed `Configuration` anyway, I'd do the same for `LoggerConfig` (embed `log-config.yaml` and parse it), otherwise it's duplicated.

ariadne/src/Ariadne/Config/Cardano.hs
```diff
@@ -1,29 +1,333 @@
...
+        , _lcShowTid  = mempty
+        , _lcConsoleAction = mempty
+        , _lcMapper = mempty
+        , _lcLogsDirectory = Nothing
+        , _lcTree = def
+        }
+
+-- TODO: define default topology value cause
+-- I didn't get which invariant is in topology.yaml
+instance Default (Topology KademliaParams) where
```
gromakovsky: Same here, it's better to embed this configuration in YAML.

ariadne/src/Ariadne/Config/CLI.hs
```diff
@@ -61,22 +62,22 @@ data CLI_NetworkConfigOpts = CLI_NetworkConfigOpts
...
     , cli_ncoBindAddress     :: !(Maybe NetworkAddress)
     , cli_ncoExternalAddress :: !(Maybe NetworkAddress)
-    } deriving (Eq, Show, Generic)
+    }
 
 data CLI_CommonArgs = CLI_CommonArgs
     { cli_logConfig            :: !(Maybe FilePath)
     , cli_logPrefix            :: !(Maybe FilePath)
     , cli_reportServers        :: !(Maybe [Text])
     , cli_updateServers        :: !(Maybe [Text])
```
gromakovsky: You deleted `reportServers` and `updateServers`, why are they still here?

iokasimov: Fair, I will check out all fields.

ariadne/src/Ariadne/Config/Cardano.hs
```diff
@@ -2,75 +2,336 @@ module Ariadne.Config.Cardano
...
+instance D.Interpret CardanoConfig where
+    autoWith _ = CardanoConfig <$> interpretCommonNodeArgs
+
+instance D.Inject CardanoConfig where
+    injectWith _ = contramap getCardanoConfig injectCommonNodeArgs
+
+defTopology :: Topology
+defTopology = either (error "LoggerConfig retrieving failed during parsing") id . decodeEither' $
+    $(embedFile $ (unsafePerformIO $ (\fp -> flip take fp $ (length fp) - 8 ) <$> getCurrentDirectory)
+        <> "/config/cardano/topology.yaml")
```
gromakovsky: It looks rather hacky and fragile. First of all, it most likely won't work on Windows, because you are using `/`. So at least `filepath` library should be used to work with paths. Second, using `unsafePerformIO` seems weird, you can use `runIO`.
I think that simpler approach is to move `config/` to `ariadne/` and embed `config/cardano/topology.yaml` and similar files. And `config/` in the root can be symlinked to `ariadne/config/`.

gromakovsky:
Superseded by #260, #269, #273, #283.

PR #246, opened 2018-07-16
[AD-300] Show error message instead of crashing on HClip exception

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-300

**Checklist:**

  - [x] [README](README.md)

**Description:**
Show error message instead of crashing on HClip exception


ui/vty/src/Ariadne/UI/Vty/Widget/Account.hs
```diff
@@ -243,5 +244,7 @@ performNewAddress = do
 performCopyAddress :: Int -> WidgetEventM AccountWidgetState p ()
 performCopyAddress idx = do
   whenJustM ((^? ix idx) <$> use accountAddressesL) $ \AccountAddress{..} -> do
-    liftIO . setClipboard . toString $ accountAddressHash
-    accountAddressResultL .= AddressCopyResultSuccess accountAddressHash
+    result <- liftIO . handle (\e -> return $ AddressResultError $ show (e :: ClipboardException)) $ do
```
Pastafarianist: You might want to use [`displayException`](https://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Exception.html#v:displayException) instead of calling `show`.

ui/vty/src/Ariadne/UI/Vty/Widget/Account.hs
```diff
@@ -4,9 +4,10 @@ module Ariadne.UI.Vty.Widget.Account
 
 import Universum
 
+import Control.Exception.Base (handle)
```
gromakovsky: Usually `Control.Exception` is imported, not `Control.Exception.Base`. To be honest, I don't know what is `Control.Exception.Base` and how it's different from `Control.Exception` :man_shrugging: 
I don't insist on using `Control.Exception`, just find `Control.Exception.Base` a bit strange.

ui/vty/src/Ariadne/UI/Vty/Widget/Account.hs
gromakovsky: Yes, `displayException` should be preferred (even if it's not explicitly defined now, it may be defined later).

ui/vty/src/Ariadne/UI/Vty/Widget/Account.hs
goodsoft: Dunno, I hoogle'd it and didn't think any more

PR #250, opened 2018-07-18
[AD-282] Allow scrolling viewport past focused widget

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-282

**Checklist:**

  - [x] [README](README.md)

**Description:**
Previously we manually used `Brick.visible` for currently focused widgets, and these "visibility requests" prevented viewports to be scrolled past these focused widgets.

Now we automatically apply `Brick.visible` on focused widgets, _except_ when `widgetIgnoreVisibility` flag is set.
This flag is set every time user scrolls viewport using mouse or `PgUp`/`PgDn`/etc buttons, and reset after every other user interaction, so that the focused widget becomes visible again.

This behaviour is automatic for all concerned widgets, except `Tree`, which is being rendered manually.

### DIFF
--- a/ui/vty/src/Ariadne/UI/Vty/Widget.hs
+++ b/ui/vty/src/Ariadne/UI/Vty/Widget.hs
@@ -26,6 +26,7 @@ module Ariadne.UI.Vty.Widget
        , setWidgetHandleEvent
 
        , getWidgetName
+       , getIgnoreVisibility
        , getFocusRing
        , findClosestFocus
        , liftBrick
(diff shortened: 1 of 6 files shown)


ui/vty/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -383,12 +391,19 @@ withWidgetState action = do
 ----------------------------------------------------------------------------
 
 drawWidget :: WidgetName -> p -> Widget p -> B.Widget WidgetName
-drawWidget focus parent (Widget widget@WidgetInfo{ widgetState, widgetDrawWithFocus }) =
-  runReader (runReaderT (widgetDrawWithFocus focus widgetState) widget) parent
+drawWidget focus parent (Widget widget@WidgetInfo{..}) =
```
maksbotan: I don't quite undestand how this works. You are adding visibility requests to all widgets, no matter focused or not?

goodsoft: No, I'm adding visibility requests only if widget is focused `(focus == widgetName)` _and_ we don't have `widgetIgnoreVisibility` flag enabled.

maksbotan: Crap, I did not notice the bit about the focus. Sorry.

PR #255, opened 2018-07-25
[AD-322] Put update URL in config and pass w/ new version event

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-322

**Checklist:**

  - [x] [README](README.md)

**Description:**
Update-related URLs are now in one config file, and no more ubiquitous `#ariadne_url` parameter, as it is passed only when needed with `UiNewVersionEvent`.
Most of the UI doesn't care about this URL at all anymore.


ariadne/src/Ariadne/Config/CLI.hs
```diff
@@ -276,7 +275,7 @@ opts :: FilePath -> Opt.ParserInfo (FilePath, Bool, CLI_AriadneConfig)
 opts xdgConfigPath = Opt.info ((parseOptions xdgConfigPath) <**> Opt.helper)
   (  Opt.fullDesc
   <> Opt.header "Ariadne wallet"
-  <> Opt.footer ("For more details see " <> toString ariadneURL))
+  <> Opt.footer ("For more details see https://serokell.io/ariadne/"))
```
goodsoft: This is the only unfortunate copy-paste of the link, but the previous "cross-backend" dependency was even worse.

gromakovsky: Maybe let's define `defaultAriadneUrl = "https://serokell.io/ariadne/"` in `Ariadne.Config.Update` and use it here and in `defaultUpdateConfig`?

goodsoft: The URL here is definitely not related to update system, and in principle shouldn't be configured at all.

gromakovsky: Imo, that's conceptually exactly the same URL as displayed by update system. Update system just tells you to go to the main website of the wallet, not to the place where it can be downloaded. Help message says the same :man_shrugging:

goodsoft: No, update system could potentially provide a direct URL to release page with version tag or something.
This one here is more of URL for a support page.

PR #258, opened 2018-07-27
[AD-323] Knit: remove Ord constraint from Knit backend

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-

**Checklist:**

  - [x] [README](README.md)

**Description:**
`Encrypted SecretKey` in Disciplina doesn't have `Ord` instance, but we still need to use it in Knit.
`Ord` instance in Knit is used only for putting type errors in `Set`, where simple `List` suffices.

### DIFF
--- a/ariadne/src/Ariadne/Knit/Backend.hs
+++ b/ariadne/src/Ariadne/Knit/Backend.hs
@@ -21,7 +21,7 @@ type Components components =
   , AllConstrained (Knit.ComponentLitToValue components) components
   , AllConstrained (Knit.ComponentInflate components) components
   , AllConstrained Knit.ComponentPrinter components
-  , Ord (Knit.Value components)
+  , Eq (Knit.Value components)
   )
 
 createKnitBackend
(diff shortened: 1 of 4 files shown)


int-index:
Now it's asymptotically slower. What about pre-rendering values as I suggested?

I don't oppose this change strongly, but I see it as a poor workaround.

goodsoft:
`Text.PrettyPrint.ANSI.Leijen.Doc` doesn't have `Ord` instance either, but `Expr CommandId components` derives its instances from `Lit`.
Well, probably I'll just convert to `Lit` then, it doesn't seem that I need non-`Ord` literals anyway.

int-index:
I don't think you can convert a `Value` to a `Lit`. Hmm. Is adding `Ord (Encrypted SecretKey)` an option?

goodsoft:
http://hackage.haskell.org/package/cryptonite-0.25/docs/Crypto-PubKey-Ed25519.html#t:SecretKey :(

Not to `Lit`, but to `Expr CommandId components`, of course.

goodsoft:
Went with a simpler approach, this is not needed for now.

PR #263, opened 2018-08-02
[DSCP-166] Make UI libs reusable and import CLI from Disciplina

goodsoft:
**YT issue:** https://issues.serokell.io/issue/DSCP-166

**Checklist:**

  - [x] [README](README.md)

**Description:**
Stack doesn't allow importing only library without executables yet.
We, however, don't want to import executables because of Cardano dependencies.


stack.yaml
```diff
@@ -9,7 +9,9 @@ packages:
 - ./ariadne/core
 - ./ariadne/cardano
 - ./ui/qt
+- ./ui/qt-app
```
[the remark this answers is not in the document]
goodsoft: Binaries will still be called `ariadne`, `ariadne-cli` and `ariadne-qt`.
Otherwise, done.

ui/qt-app/ariadne-qt-app.cabal
```diff
@@ -0,0 +1,79 @@
+name:           ariadne-qt-app
+version:        0.2.0.0
+description:    Please see the README on Github at <https://github.com/serokell/ariadne#readme>
```
gromakovsky: Now that we have many packages, I think it'd be good to write something more concrete here. Not necessary in this PR, just speaking loudly.

ui/qt-app/ariadne-qt-app.cabal
```diff
@@ -0,0 +1,79 @@
...
+      TemplateHaskell
+      TupleSections
+      TypeApplications
+      TypeFamilies
+      TypeOperators
+      UndecidableInstances
+      ViewPatterns
+      EmptyCase
+      PartialTypeSignatures
+  ghc-options: -Wall -Wcompat -Werror
```
gromakovsky: I noticed that there is `ghc-options:` on line 20 and then here. Wouldn't it be better to merge them?

ui/vty-app/ariadne-vty-app.cabal
```diff
@@ -0,0 +1,81 @@
...
+      TemplateHaskell
+      TupleSections
+      TypeApplications
+      TypeFamilies
+      TypeOperators
+      UndecidableInstances
+      ViewPatterns
+      EmptyCase
+      PartialTypeSignatures
+  ghc-options: -Wall -Wcompat -Werror
```
gromakovsky: For this executable `ghc-options` is duplicated as well. I don't insist on merging them in this PR, but consider doing it while you're at it if you agree it's a good idea.

ui/cli-app/ariadne-cli-app.cabal
```diff
@@ -0,0 +1,75 @@
...
+    , ariadne-cardano
+    , ariadne-cli-lib
+    , ariadne-core
+    , async
+    , base >=4.7 && <5
+    , ii-extras
+    , knit
+    , text
+    , universum
+  other-modules:
```
gromakovsky: `other-modules` can be merged, too.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -48,6 +48,12 @@ data RenameResult
   | RenameResultError !Text
   | RenameResultSuccess
 
+data BalanceResult
```
gromakovsky: Why is it copy-pasted? Isn't there a better solution?

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
```diff
@@ -46,6 +48,15 @@ import Data.Tree (Tree)
 import Data.Version (Version)
 import Text.PrettyPrint.ANSI.Leijen (Doc)
 
+-- | UI library settings for a particular currency implementatio
```
gromakovsky: implementation

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Wallet.hs
goodsoft: I consider this widget-local structure.
Other way would be defining polymorphic result type `Result a` and using it for all requests (as they only differ by result type), but that's for another PR.

ui/vty-app/ariadne-vty-app.cabal
goodsoft: I'm not particularly fluent in Cabal, so I don't have any strong opinion on this.
And I think this should be resolved along with hpack integration.

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
gromakovsky: Ok, just fix this typo then.

PR #269, opened 2018-08-10
[AD-336] Permit static values for Configuration, LogConfig, Topology

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-336

**Checklist:**

  - [x] [README](README.md)

**Description:**

_Prehistory_
If `--config` option (which defaults to `xdgConfigPath </> "ariadne-config.dhall"`) points to a non-existing file, the default configuration will be used. It can happen for users who built `ariadne` from source, launched it, but didn't copy/symlink configuration into default location. The default configuration uses filepaths like `"config/cardano/topology.yaml"`. That's not ideal because these files might not exist (it will most likely happen if someone launches `ariadne` not from the root of the repository). Ideally we want `ariadne` to be usable regardless of presence of any configuration files on disk.

There are three configuration files which Cardano uses. One of them is mandatory (configuration). It refers to another file, so in some sense there are four files. Other two are optional, but their default values are defined in `cardano-sl` and we want them to be different.
There are (at least) three possible ways to use static values instead of reading data from files:
1. Modify code in `cardano-sl` to accept a default value. Works only for optional files. If the file is mandatory, it should be made optional first.
2. Copy-paste function which uses a default value and modify it in our code. Works only for optional files. If the file is mandatory, it should be made optional first.
3. Create necessary file, dump data there and use its path. Works even for mandatory files.

_Solution_
I spent quite a lot of time on this issue and somehow ended up using all three approaches.
Default logger config was specified in the `readLoggerConfig` function. This function is trivial, so I just copy-pasted it in our code, so that now default logger config is defined in our code.
Default topology was defined in `Pos.Infra.Network.CLI` and it was rather deep. I used combination of (1) and (2). `intNetworkConfigOpts` now accepts default topology. It's used by `getNodeParams` which I copied in our code.
For `Configuration` I decided to use option (3), because otherwise I would have to copy-paste too much code (and modify it in `ariadne`) or generalize too much code directly in `cardano-sl`.

### DIFF
--- a/ariadne/cardano/src/Ariadne/Cardano/Backend.hs
+++ b/ariadne/cardano/src/Ariadne/Cardano/Backend.hs
@@ -4,6 +4,7 @@ import Universum
 
 import Control.Concurrent.STM.TVar (TVar)
 import Control.Monad.Trans.Reader (withReaderT)
+import qualified Data.ByteString as BS
 import Data.Constraint (Dict(..))
 import Data.Maybe (fromJust)
 import Data.Ratio ((%))
@@ -18,7 +19,9 @@ import Pos.DB.DB (initNodeDBs)
 import Pos.Infra.Diffusion.Types (Diffusion, hoistDiffusion)
 import Pos.Infra.Slotting (MonadSlots(getCurrentSlot, getCurrentSlotInaccurate))
 import Pos.Launcher
-import Pos.Launcher.Configuration (withConfigurations)
+  (ConfigurationOptions(cfoFilePath), LoggingParams(..), NodeParams(..),
+  runNode, runRealMode)
+import qualified Pos.Launcher.Configuration as Launcher.Configuration
 import Pos.Launcher.Resource (NodeResources(..), bracketNodeResources)
 import Pos.Txp (txpGlobalSettings)
 import Pos.Update.Worker (updateTriggerWorker)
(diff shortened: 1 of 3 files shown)


ariadne/cardano/src/Ariadne/Cardano/Backend.hs
```diff
@@ -111,10 +118,17 @@ runCardanoNode protocolMagic bHandle addUs cardanoContextVar diffusionVar
         convertMode f diff =
             cardanoModeToRealMode $ f (hoistDiffusion realModeToCardanoMode cardanoModeToRealMode diff)
 
+        -- It's needed because 'bracketNodeResources' uses its own
+        -- logic to create a logging config, which differs from what
+        -- we do above. We want 'ncLoggerConfig' to be the same config
+        -- as the one passed to 'setupLogging'.
+        setProperLogConfig :: NodeResources __ -> NodeResources __
```
Pastafarianist: Why do you need double underscores here?

ariadne/cardano/src/Ariadne/Cardano/Backend.hs
```diff
@@ -111,10 +118,17 @@ runCardanoNode protocolMagic bHandle addUs cardanoContextVar diffusionVar
...
+        setProperLogConfig nr =
+            nr {nrContext = (nrContext nr) {ncLoggerConfig = loggerConfig}}
```
Pastafarianist: Are there no lenses for `NodeResources` and `NodeContext`? Using them here would seem like a much cleaner solution.

PR #273, opened 2018-08-13
AD-345 Embed static configs and use them by default 

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-345

**Checklist:**

  - [x] [README](README.md)

**Description:**
Default logging config and topology are now embedded into `ariadne` binary, as well as Cardano's `Configuration`.
Default Ariadne configuration now omits 'log-config' and 'topology', embedded values are used for them instead of file paths.
Cardano-specific configuration files have been moved to `ariadne/cardano/`, because they are details of Cardano and otherwise it would be hard to embed them.
`defaultAriadneConfig` and `config/ariadne-config.dhall` are now identical¹, I think it's good, previously there was some confusion because we had more than one _default_ config.
I'd like to keep both in the repository. In Haskell code we need `defaultCommonNodeArgs` 
anyway. And we need to have some sort of a template for people who want to modify configuration.

¹ Not fully, see my second commit message, it will be dealt with in AD-350.


gromakovsky:
>  TODO: I should probably add a test that they are identical.

Upd: I decided not to do it now, let's do it after project structure becomes more granular (currently there is too much stuff in `ariadne/cardano`) and configuration approach also becomes more modular (AD-338).

> LGTM, but does not build for some reason.

Hm, that's probably because of the bug I saw earlier: if there is a dependency in tests which is not a dependency in library, CI fails for some reason. Will try to workaround.

> Are you going to write READMEs about cardano configs, btw?

I think it's a good thing to do indeed. Will create an issue.

PR #275, opened 2018-08-14
[AD-229] Use componentm for components creation

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-229

**Checklist:**

  - [x] [README](README.md)

**Description:**
`componentm` has some generally useful features, but the main reason
for us to use it is that it should make it much easier to use `bracket`
style initialization of components. However, this commit doesn't add
cleanup actions, it only refactors existing code. Cleanup actions will
be added later. So only `buildComponent_` is used now, but `buildComponent`
will also be used.
Common functionality from `Main.hs` modules was moved into a new module called
`MainTemplate` because I didn't want to make the same changes in three places.

### DIFF
--- a/ariadne/cardano/src/Ariadne/MainTemplate.hs
+++ b/ariadne/cardano/src/Ariadne/MainTemplate.hs
@@ -1,0 +1,133 @@
+-- | Template of the code that is supposed to be in 'Main.hs'.
+
+module Ariadne.MainTemplate
+       ( MainSettings (..)
+       , defaultMain
+       ) where
+
+import Universum
+
+import Control.Concurrent.Async (race_, withAsync)
+import Control.Monad.Component (ComponentM, runComponentM)
+import Data.Version (Version)
+import Data.Vinyl.Core (Rec(..), (<+>))
+import Data.Vinyl.TypeLevel (type (++))
+import IiExtras (runNat)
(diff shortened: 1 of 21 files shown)


ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -0,0 +1,133 @@
...
+                                   , acUpdate = updateConfig
+                                   , acHistory = historyConfig
+                                   } = do
+  history <- openCommandHistory $ hcPath historyConfig
+
+  (uiFace, mkUiAction) <- msCreateUI history
+  (bHandle, addUs, mkWallet) <-
+      createWalletBackend walletConfig (msPutWalletEventToUI uiFace)
+  (cardanoFace, mkCardanoAction) <- createCardanoBackend cardanoConfig bHandle addUs
+  let CardanoFace { cardanoRunCardanoMode = runCardanoMode
```
maksbotan: Why not just `let runCardanoMode = cardanoRunCardanoMode cardanoFace`?

gromakovsky: I don't know, that's how it was in `Main.hs` files and I didn't want to change it.

PR #276, opened 2018-08-15
[AD-269] Show wallet ids in TUI

maksbotan:
**YT issue:** https://issues.serokell.io/issue/AD-269

**Checklist:**

  - [x] [README](README.md)
- [X] Adressed HLint warnings and hints
- [X] Tested my changes if they modify code

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/UiAdapter.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/UiAdapter.hs
@@ -42,7 +42,7 @@ import Control.Lens (makeLenses)
 -- 'WalletData' like data type, used only in UI glue
 data UiWalletData = UiWalletData
   { _uwdName     :: !Text
-  , _uwdId       :: !Text
+  , _uwdId       :: !HdRootId
   , _uwdAccounts :: !(Vector UiAccountData)
   , _uwdBalance  :: !Coin
   } deriving (Show, Generic)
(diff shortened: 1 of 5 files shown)


ariadne/cardano/src/Ariadne/Wallet/UiAdapter.hs
```diff
@@ -34,14 +34,15 @@ import Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet
...
+import Pos.Crypto.Hashing (AbstractHash(..))
 import Serokell.Util (enumerate)
 
 import Control.Lens (makeLenses)
 
 -- 'WalletData' like data type, used only in UI glue
 data UiWalletData = UiWalletData
   { _uwdName     :: !Text
-  , _uwdId       :: !HdRootId
+  , _uwdId       :: !Text
```
maksbotan: This field was not used anywhere. Alternative is formatting hash in `Glue`, but that does not depend on cardano-sl-crypto, so I can't import `Pos.Crypto.Hashing` to pattern match `AbstractHash`. I don't think we want cardano dependency in Glue.

goodsoft: `Glue` is part of `ariadne-*-app`, which depends on `ariadne-cardano` anyway.
IMHO, backend should spew out as typed values as possible, while frontend should take as generic values as possible.
Glue connects them according to the concrete executable needs.

ariadne/cardano/src/Ariadne/Wallet/UiAdapter.hs
```diff
@@ -97,7 +98,7 @@ toUiWalletDatas db = toUiWalletData <$> walletList
     toUiWalletData :: HdRoot -> UiWalletData
     toUiWalletData HdRoot{..} = UiWalletData
       { _uwdName = unWalletName _hdRootName
-      , _uwdId = _hdRootId
+      , _uwdId = let (AbstractHash x) = _fromDb $ unHdRootId $ _hdRootId in show x
```
gromakovsky: I think that using `sformat hashHexF` would be better, because it makes it clearer that hex format will be used (and because that's how `componentTokenRender` is defined for `TokenAddressHash`).

goodsoft: It's even weirder now.
Better define one function in `ariadne-cardano`, which both `Glue` and `Knit` will use.

PR #277, opened 2018-08-15
[AD-273] Delete whole subtree when something is deleted. 

Haskell-mouse:
**YT issue:** https://issues.serokell.io/issue/AD-273

**Checklist:**

  - [ ] [README](README.md)

**Description:**

Added deletion of the whole subtree (when someone delete wallet or account). 

For that I added separate function for deletion lists of acoounts/addresses.
I think, that it should be slightly more effective due to less lens operations,
but if it does not matter i just can use mapM_ with single-value functions in KeyStorage.hs and delete new functions
from delete.hs.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,11 +1,12 @@
 let
-  overlay = import ''${builtins.fetchGit "ssh://git@github.com/serokell/serokell-overlay"}/pkgs'';
-  nixpkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/master.tar.gz") {
-    overlays = [ overlay ];
-  };
+  nixpkgs = fetchTarball "https://github.com/serokell/nixpkgs/archive/master.tar.gz";
+  serokell-overlay = fetchGit "ssh://git@github.com/serokell/serokell-overlay";
 in
 
-with nixpkgs;
+with import nixpkgs {
+  config.allowUnfree = true;
+  overlays = [ (import "${serokell-overlay}/pkgs") ];
+};
 
 let
   closure = (stackClosure haskell.compiler.ghc822 ./.).override {


ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -527,15 +527,25 @@ removeSelection
...
     -- Throw "Nothing selected" here?
     Just selection -> case selection of
       WSRoot hdrId -> do
+        let 
+          accList = (getAccounts (walletDb ^. dbHdWallets) hdrId) <&> (^. hdAccountId)
+          addrList = (getAddressesByRoot (walletDb ^. dbHdWallets) hdrId) <&> (^. hdAddressId)
+        -- delete the whole subtree. 
+        update acidDb (DeleteHdAddresses addrList)
+        update acidDb (DeleteHdAccounts accList)
         update acidDb (DeleteHdRoot hdrId)
```
gromakovsky: This is wrong, there should be only one `update`, because we want to delete the whole tree atomically. One `update` is done atomically, but three `update`s are not.
You should defined something like `DeleteHdTree` in `Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Delete` which will delete whole tree.
Something similar should be done for account deletion as well (there are two `update`s currently).

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -544,6 +554,20 @@ removeSelection acidDb WalletFace{..} walletSelRef runCardanoMode = do
...
+      (readAccountsByRootId rootId wallets)
+
+    getAddressesByRoot :: HdWallets -> HdRootId -> [HdAddress]
+    getAddressesByRoot wallets rootId = toList $ fromRight 
+      (error "Bug: UnknownHdRoot")
+      (readAddressesByRootId rootId wallets) 
+
+    getAddressesByAccount :: HdWallets -> HdAccountId -> [HdAddress]
+    getAddressesByAccount wallets accId = toList $ fromRight
+      (error "Bug: UnknownHdAccount")
```
gromakovsky: These three `error`s look unpleasant and can be avoided if you move this logic into `Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Delete` (there you can delete all addresses corresponding to given `AccountId`).

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet/Delete.hs
```diff
@@ -10,15 +12,38 @@ import Control.Lens (at, (.=))
...
+
 
--- | Delete an account
-deleteHdAccount :: HdAccountId -> Update' HdWallets UnknownHdRoot ()
-deleteHdAccount accId = zoom hdWalletsAccounts $ at accId .= Nothing
+-- | Delete an account with it's addresses
+deleteHdAccount :: HdWallets -> HdAccountId -> Update' HdWallets UnknownHdRoot ()
+deleteHdAccount hdwallets accId = do 
+  let addrIds = (getAddressesByAccount hdwallets accId) <&> (^. hdAddressId)
+  zoom hdWalletsAddresses $ mapM_ (\addrId -> (at addrId .= Nothing)) addrIds
```
gromakovsky: Here and in `deleteHdRoot` we basically want to delete all items from `IxSet` with given index. So I have a cosmetic suggestion: define a function with type `(Indexable a, IsIndexOf ix a) => ix -> IxSet a -> IxSet a` in `Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet` and use it here and above. It allows to avoid usage of `error` and reduces code duplication.
Possible definition (I didn't test it):
```
deleteIxAll ix (WrapIxSet ixSet) =
    WrapIxSet (IxSet.getGT ix ixSet `IxSet.union` IxSet.getLT ix ixSet)
```
Note: there is `deleteIx` in `ixset-typed`, but it can't be used, because it
> Will not change IxSet if you have more than one item with given index.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet/Delete.hs
```diff
@@ -6,19 +8,46 @@ module Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Delete (
...
+  checkExistence $ AccountId accId
+  zoom hdWalletsAddresses $ modify (deleteIxAll accId)
+  zoom hdWalletsAccounts  $ modify (deleteIxAll accId) 
+
 
--- | Delete an account
-deleteHdAccount :: HdAccountId -> Update' HdWallets UnknownHdRoot ()
-deleteHdAccount accId = zoom hdWalletsAccounts $ at accId .= Nothing
+checkExistence :: Selected a -> Update' HdWallets (UnknownId a) ()
+checkExistence (RootId rootId) =
```
gromakovsky: It seems to me that this function is unnecessary complication. Its body is defined for each case separately, so you could instead define two functions and their definitions would take the same amount of code. The only thing it buys is that you don't need to type its signature twice, but instead you need to define an extra data type and a type family.

gromakovsky:
Why did you close it?

PR #283, opened 2018-08-17
[AD-350] Make two defaults configs identical

gromakovsky:
**YT issue:** https://issues.serokell.io/issue/AD-350

**Checklist:**

  - [x] [README](README.md)

**Description:**
Previously there was a small difference between using `config/ariadne-config.dhall` and using hardcoded config (which is used when config file can't be found), because relative paths in hardcoded config weren't treated as relative to 'configPath', while paths in
config file were.
Having two different default configs seems bad and confusing, so this difference was eliminated.
The only relative path in the default config is path to `cardano-configuration.yaml`. In default case the only difference for a user is that this file (and `mainnet-genesis.json`) will be dumped into `~/.config/ariadne/` instead of current directory, which is better, I think.

### DIFF
--- a/ariadne/cardano/src/Ariadne/Config/CLI.hs
+++ b/ariadne/cardano/src/Ariadne/Config/CLI.hs
@@ -153,20 +153,19 @@ getConfig commitHash = do
     exitSuccess
 
   xdgDataPath <- getXdgDirectory XdgData "ariadne"
-  config <- ifM (doesFileExist configPath)
-    (do
-      -- Dhall will throw well formatted colourful error message
-      -- if something goes wrong
+  configDirs <- ConfigDirectories xdgDataPath <$> getCurrentDirectory
+  unresolvedConfig <- ifM (doesFileExist configPath)
+    -- Dhall will throw well formatted colourful error message
+    -- if something goes wrong
 
-      -- Passing path as dhall import is needed for relative import paths
-      -- to be relative to the config path.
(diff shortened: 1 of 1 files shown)


ariadne/cardano/src/Ariadne/Config/CLI.hs
```diff
@@ -153,20 +153,19 @@ getConfig commitHash = do
...
+  unresolvedConfig <- ifM (doesFileExist configPath)
       -- Dhall will throw well formatted colourful error message
       -- if something goes wrong
 
       -- Passing path as dhall import is needed for relative import paths
       -- to be relative to the config path.
-      unresolved <- fromDhall @AriadneConfig $ toDhallImport configPath
-      configDirs <- ConfigDirectories xdgDataPath <$> getCurrentDirectory
-      return (resolvePaths unresolved configPath configDirs))
+      (fromDhall @AriadneConfig $ toDhallImport configPath)
```
goodsoft: Indentation got all weird here

gromakovsky:
> unlike the rest of the configuration logic

Do you mean dumping `cardano-configuration.yam` and `mainnet-genesis.json` on disk or something else?

goodsoft:
I mean the fact that `Ariadne.Config.CLI` is such a _telega_ and includes Cardano-specific stuff.
Not related to this particular PR, just ranting.

PR #288, opened 2018-08-23
[AD-326] Port IOHK updates

Pastafarianist:
**YT issue:**
* https://issues.serokell.io/issue/AD-326
* https://issues.serokell.io/issue/AD-306
* https://issues.serokell.io/issue/AD-176

**Checklist:**

- [x] Updated docs if necessary _(not necessary)_

**Description:**

This is basically a massive refactoring. The only change that should be visible on the outside is that the path to key storage is now passed as `--wallet:keyfile` instead of `--cardano:keyfile`.

Summary of changes:

* `[AD-306]` `WalletLayer` now sits in between of (most) calls between the `Kernel` and the rest of the code. The general pattern is that we create a `PassiveWalletLayer` in `Wallet.Backend` and pass it to various functions, which are mostly defined in `KeyStorage`. These functions then call methods from the `PassiveWalletLayer` instead of invoking `Kernel` functions directly. The implementations are mostly in `Kernel.{Wallets,Accounts,Addresses}`. The interface and wrappers are in `WalletLayer.{Types,Kernel}`.
* `addUs` and the whole hackery around it is gone. Now wallet manages its own `Keystore`. The original implementation by IOHK does not persist changes at each update (it only saves them at shutdown), which was fixed, even though IOHK have an [issue](https://github.com/input-output-hk/cardano-sl/blob/44c903587d453cea86dae4ef8549430e34b2449f/wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs#L79) registered for that.
* The path to key storage is now passed as `--wallet:keyfile` instead of `--cardano:keyfile`. The latter parameter is no longer supported (Ariadne will throw an error if it is passed).
* `BlockMeta` is now hidden within a `PrefilteredBlock`.
* `Kernel.DB.HdWallet.Derivation` is a new collection of helper functions, mostly from `KeyStorage`.
* `HasSpendingPassword` is no longer passed around, instead it is generated in `Kernel.Wallets`. Instead, we use `HasNonemptyPassword`, which is a `newtype` over `Bool`.
* `WalletLayer.Kernel` now exports `passiveWalletLayerWithDBComponent` in addition to `passiveWalletLayerComponent`; the former allows to specify a custom Acid DB instead of creating one in-memory.
* `import_wallet.sh` is updated.

Substantial differences with upstream are:

* no `CoinSelection`
* no `MonadDBReadAdapter`
* instead of upstream implementation of BIP-39 we're using the one from Lootbox
* `Derivation` is completely different
* `HdRootId` wraps `AddressHash PublicKey` instead of `Address`
* `PrefilteredBlock` is very different
* no limits on execution time of API functions
* `ComponentM` instead of `bracket`

Remaining TODOs:

- [x] Compare old implementations of various functions in `KeyStorage` and new ones in `{Wallets,Accounts,Addresses}` and make sure nothing is forgotten.
- [x] Make `Keystore` persist data on each update.
- [x] Use `usWallets` instead of `usKeys` in `Keystore`.
- [x] Remove everything we aren't using.
- [x] Refactor wallet/account name autogeneration: move account name generation from `Kernel.DB.HdWallet.Create.initHdAccount` into `KeyStorage` and make it uniform with the one used for wallet names.
- [x] There is now a single function for changing wallet parameters, `pwlUpdateWallet`. It accepts both a new name and a new assurance and does not allow to modify them separately. Need to split it into multiple functions.
- [x] Register an issue for wiring `newPending` into `sendTx`. (AD-369)
- [x] Register an issue about introducing `bracket*` functions for test compatibility. They can be made from `ComponentM` stuff using `runComponentM`.
- [x] Register an issue about abstracting away the existence of AcidState DB in `PassiveWalletLayer`. Currently it exposes `pwlGetDBSnapshot`. This should probably be fixed by exposing more specialized functions. (AD-370)
- [x] Register an issue about optimizing sequential index derivation by storing the necessary metadata with `WalletIdHdSeq` and `AccountIdHdSeq`. (AD-371)
- [x] Register an issue for adopting stuff that is omitted here:
    * `ActiveWallet`
    * submission layer
    * coin selection
 (Filed as AD-256.)
- [x] `import_wallet.sh` now contains a very long line that I would like to see shortened, but I don't know enough Bash to do that.
- [x] Replace `Rnd` with `Seq` in wallet/account/address generation function names. Currently they still contain `Rnd` even though they no longer generate things randomly, as they were in IOHK implementation.
- [x] Fix outdated comments.
- [x] Run `hlint` on this.
- [x] Test.

As a final note, it seems that our vision differs too much with that of IOHK. In particular:

* we use BIP-44 and the whole `{Acc,Addr}Checkpoint` dichotomy
* we use `Map HdAccountId PrefilteredUtxo` as a format for passing prefiltered utxo around, and our definition of `PrefilteredUtxo` is different (more structured)
* we use `AddressHash PublicKey` as a root key identifier instead of a more malleable `Address`
* we don't need to expose an API that can be served via HTTP (and thus don't use anything from `V1.Types`)
* we don't set limits on the time for wallet API method execution

so at this moment I'm not sure how, when and whether further imports will proceed after we merge the [tests](https://github.com/serokell/ariadne/compare/thatguy/ad167-port-iohk-tests-0fa0ce6?expand=1).


ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -482,106 +354,70 @@ select acidDb WalletFace{..} walletSelRef mWalletRef uiPath = do
...
   -> IO ()
-renameSelection acidDb WalletFace{..} walletSelRef name = do
+renameSelection pwl WalletFace{..} walletSelRef name = do
   mWalletSel <- readIORef walletSelRef
   case mWalletSel of
     Nothing -> pure ()
     Just selection -> case selection of
       WSRoot hdrId ->
-        throwLeftIO $ update acidDb (UpdateHdRootName hdrId (WalletName name))
+        -- TODO: do not hardcode assurance level
```
maksbotan: maybe add task id here as well?

ariadne/cardano/src/Ariadne/Wallet/Backend/Restore.hs
```diff
@@ -117,20 +116,20 @@ restoreFromKeyFile acidDb face runCardanoMode mbWalletName path rType = do
...
+    -> HasNonemptyPassphrase
     -> AssuranceLevel
     -> IO ()
-restoreFromSecretKey acidDb face runCardanoMode mbWalletName esk rType hasPass assurance = do
+restoreFromSecretKey pwl face runCardanoMode mbWalletName esk rType hasPP assurance = do
     utxoByAccount <- case rType of
         WalletRestoreQuick -> pure mempty
         WalletRestoreFull  -> runCardanoMode $ collectUtxo esk
-    addWallet acidDb face runCardanoMode esk mbWalletName utxoByAccount hasPass assurance
+    addWallet pwl face esk mbWalletName utxoByAccount hasPP assurance
```
maksbotan: would it make some sense to turn all these functions into `ReaderT PassiveWalletLayer IO`? that'd increase the diff though...

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
```diff
@@ -87,37 +91,68 @@ import Ariadne.Wallet.Cardano.Kernel.Word31 (Word31)
   Supporting types
 -------------------------------------------------------------------------------}
 
--- | Name of a wallet.
+-- | Wallet name
```
maksbotan: Great change :D

ariadne/cardano/src/Ariadne/Wallet/Backend.hs
```diff
@@ -41,12 +40,16 @@ createWalletBackend walletConfig sendWalletEvent = do
...
 
-    (us :: TVar UserSecret, addUs :: TVar UserSecret -> IO ()) <- newInitFuture "UserSecret"
-    PassiveWalletLayer{..} <- passiveWalletLayerComponent
-        (usingLoggerName "passive-wallet" ... logMessage) us acidDb
-    let refresh = refreshState acidDb walletSelRef sendWalletEvent
-        applyHook = const refresh <=< _pwlApplyBlocks
-        rollbackHook = const refresh <=< _pwlRollbackBlocks
+    -- TODO: configurable path to keyfile
+    keystore <- keystoreComponent RemoveKeystoreIfEmpty "secret-mainnet.key"
+    pwl <- passiveWalletLayerWithDBComponent
```
gromakovsky: Maybe let's not use `acidDB` outside wallet layer at all? And then create it inside `PassiveWalletLayer` creation function and not expose outside? It should be trivial currently, because it's used only in two places here and `_pwlGetDBSnapshot` is enough.
It will highlight that all code which deals with `acid-state` is hidden behind `PassiveWalletLayer`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Actions.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
```
gromakovsky: Enabled by default.

ariadne/cardano/src/Ariadne/Wallet/Backend/Restore.hs
gromakovsky: Definitely not in this PR.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
Pastafarianist: https://github.com/input-output-hk/cardano-sl/blob/0fa0ce646cb83af8028878a1eed169f513a987b6/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs#L90

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -253,9 +356,10 @@ createPrefiltered :: forall p e.
                   -> Map HdAccountId AccountName
                   -> Update' HdWallets e ()
 createPrefiltered mkPrefilteredUtxo accApplyP narrowP addrApplyP pByAccount accountNames =
-    forM_ (Map.toList pByAccount) $ \(accId, p) ->
-        createAccPrefiltered mkPrefilteredUtxo accApplyP narrowP addrApplyP accId p
+    forM_ (Map.toList pByAccount) $ \(accId, p) -> do
+        _ <- createAccPrefiltered mkPrefilteredUtxo accApplyP narrowP addrApplyP accId p
```
gromakovsky: That's most likely not necessary and imo it doesn't make anything better :man_shrugging:

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
```diff
@@ -87,37 +91,68 @@ import Ariadne.Wallet.Cardano.Kernel.Word31 (Word31)
...
 newtype WalletName = WalletName
     { unWalletName :: Text
     } deriving (Show, Eq, Ord, IsString, ToString)
 
+instance Buildable WalletName where
+    build (WalletName wName) = bprint build wName
```
gromakovsky: Could have used GND instead :man_shrugging:

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
```diff
@@ -440,8 +535,8 @@ zoomOrCreateHdRoot :: HdRoot
                    -> HdRootId
                    -> Update' HdRoot    e a
                    -> Update' HdWallets e a
-zoomOrCreateHdRoot newRoot rootId  =
-    zoomCreate newRoot (hdWalletsRoots . at rootId)
+zoomOrCreateHdRoot newRoot rootId upd =
+    zoomCreate newRoot (hdWalletsRoots . at rootId) $ upd
```
gromakovsky: :man_shrugging:

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
```diff
@@ -16,29 +16,47 @@ module Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet (
...
+    -- * Destruction
   , toAscList
+  , toList
     -- * Indexing
   , (@+)
   ) where
 
-import Universum
+import qualified Prelude
+import Universum hiding (Foldable, null, toList)
```
gromakovsky: Hiding `Foldable` is useless, let's not do it.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
```diff
@@ -157,6 +181,27 @@ member pk = isJust . view (Lens.at pk)
 size :: IxSet a -> Int
 size = IxSet.size . unwrapIxSet
 
+-- | Whether or not this 'IxSet' contains no elements.
+null :: IxSet a -> Bool
+null = IxSet.null . unwrapIxSet
```
gromakovsky: I'd rather define `null = Foldable.null` or `null = Container.null` or `Foldable.null = null`. Same for `toList`.

(I know it will increase diff with IOHK, but I think we should stop code imports after this one)

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
```diff
@@ -165,7 +210,7 @@ updateIxManyM
     :: (Indexable a, IsIndexOf ix a, Applicative f)
     => ix -> (a -> f a) -> IxSet a -> f (IxSet a)
 updateIxManyM pk f (WrapIxSet s) = do
-    let originalValues = toList $ IxSet.toSet $ IxSet.getEQ pk s
+    let originalValues = Data.Foldable.toList $ IxSet.toSet $ IxSet.getEQ pk s
```
gromakovsky: So much hate :unamused:

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Read.hs
```diff
@@ -0,0 +1,90 @@
+module Ariadne.Wallet.Cardano.Kernel.DB.Read (
```
gromakovsky: This module is weird and likely useless.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
Pastafarianist: Do you mean using `_ <- ...; pass` instead of `void $` or returning the newly created `HdAccount` from `createAccPrefiltered`?

gromakovsky: You don't need `void`, `forM_` should discard results for you.

gromakovsky: `forM_ :: (Foldable t, Monad m) => t a -> (a -> m b) -> m ()`

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
Pastafarianist: :man_shrugging:

Pastafarianist: The point here is that there are `zoomOrCreateHd{Account,Address}` where you cannot reduce `upd` similarly, so this change makes their implementations more uniform. But I see no point in arguing over this: it is super minor, and if you insist, I'll revert this.

gromakovsky: > it is super minor

Yes, it is, that's why I wrote :man_shrugging: instead of "wtf fix it pls".

> if you insist

I don't.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
Pastafarianist: I suppose we can just remove this function since it is defined in `Foldable` anyway.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
Pastafarianist: Hope you will like the newer version.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -0,0 +1,173 @@
...
+    case res of
+         Left e   -> return . Left $ CreateWalletFailed e
+         Right hdRoot -> do
+             -- STEP 2: Insert the 'EncryptedSecretKey' into the 'Keystore'
+             Keystore.insert (WalletIdHdSeq newRootId) esk (pw ^. walletKeystore)
+             return (Right hdRoot)
+
+
+-- | Creates an HD wallet where new accounts and addresses are generated
+-- via random index derivation.
```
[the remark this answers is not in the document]
Pastafarianist: Thanks. Didn't notice that.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Resolved.hs
```diff
@@ -8,19 +8,22 @@ module Ariadne.Wallet.Cardano.Kernel.DB.Resolved (
...
+  , rbSlot
   ) where
 
 import Universum
 
 import Control.Lens.TH (makeLenses)
 import qualified Data.Map as Map
 import Data.SafeCopy (base, deriveSafeCopySimple)
-import qualified Data.Text.Buildable
+import Data.Text.Buildable
```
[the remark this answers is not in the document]
Pastafarianist: No idea why they did that :man_shrugging:

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -387,89 +278,62 @@ generateMnemonic entropySize = do
...
-  -- advanced feature and do not provide it for now.
-  -- generateMnemonic function guarantees that mnemonic is a non-empty list
-  let seed = mnemonicToSeed (unwords $ Unsafe.init mnemonic) ""
-  let (_, esk) = safeDeterministicKeyGen seed pp
-  hasPass <- mkHasPass runCardanoMode (pp == emptyPassphrase)
-  mnemonic <$
-    addWallet acidDb face runCardanoMode esk mbWalletName mempty hasPass assurance
+  let seed = mnemonicToSeedNoPassword $ unwords $ Unsafe.init mnemonic
+      (_, esk) = safeDeterministicKeyGen seed pp
+  mnemonic ++ ["ariadne-v0"] <$
```
maksbotan: In `generateMnemonic` I already append `"ariadne-v0"`. Please sync this code with master.

PR #290, opened 2018-08-26
[AD-365] Remove IiExtras

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-365

**Checklist:**

  - [ ] [README](README.md)

**Description:**
Inspired by #93.


goodsoft:
Regarding custom prelude, I think it's moderately fine to have different conventions in different _packages_.

I'm actually more concerned about amount of stuff `Knit` exports right now, as typing `hiding` lists everywhere we use Knit does not seem adequate to me.

int-index:
> `Knit` shouldn't re-export its prelude.

OK, but it definitely should re-export some parts of it. For example, `uprism` is defined there, and if `uprism` is exported, then it makes sense to re-export the entirety of `ntype`.

> Also @int-index promised to upload ntype, let's wait for that to happen.

Don't wait for this, I will not upload before I write some documentation, and I'm not writing documentation any time soon.

int-index:
>  I will not upload before I write some documentation

Now I remember our conversation where you convinced me to upload even without documentation. OK, I will do it in a moment.

int-index:
Uploaded https://hackage.haskell.org/package/ntype

PR #291, opened 2018-08-27
[AD-262] Add TUI viewport with scrollbar

pasqu4le:
**YT issue:** https://issues.serokell.io/issue/AD-262

**Checklist:**

  - [ ] [README](README.md)

### DIFF
--- a/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
+++ b/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
@@ -78,7 +78,7 @@ fixedViewport name vpType p = case vpType of
   B.Horizontal -> B.Widget B.Greedy (B.vSize p) render
   where
     render = do
-      result <- renderReleased vpType p
+      result <- B.render p
       B.render $
         limit result .
         B.viewport name vpType $
--- a/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
+++ b/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
@@ -117,9 +117,12 @@ viewportWithScrollBar name vpType p = B.Widget B.Greedy B.Greedy $ do
             B.Widget (B.hSize p) (B.vSize p) (return result)
 
         resultImg = B.image result
-
-        enoughHeight = V.imageHeight resultImg <= B.availHeight c
-        availHeight = fromIntegral $ B.availHeight c
+        (availH, availW) = case vpType of
+            B.Both -> (B.availHeight c - 1, B.availWidth c - 1)
+            _ -> (B.availHeight c, B.availWidth c) 
+    
+        enoughHeight = V.imageHeight resultImg <= availH
+        availHeight = fromIntegral availH
         resultHeight = fromIntegral $ V.imageHeight resultImg
         vpPadTop = fromIntegral $ maybe 0 (view B.vpTop) mvp
         vRatio = availHeight / resultHeight :: Double
@@ -128,8 +131,8 @@ viewportWithScrollBar name vpType p = B.Widget B.Greedy B.Greedy $ do
         vBar = B.vLimit (vPad + vSize) $ B.padTop (B.Pad vPad) BR.vBorder
         addVBar widg = if enoughHeight then widg else B.hBox [widg, vBar]
 
-        enoughWidth = V.imageWidth resultImg <= B.availWidth c
-        availWidth = fromIntegral $ B.availWidth c
+        enoughWidth = V.imageWidth resultImg <= availW
+        availWidth = fromIntegral availW
         resultWidth = fromIntegral $ V.imageWidth resultImg
         vpPadLeft = fromIntegral $ maybe 0 (view B.vpLeft) mvp
         hRatio = availWidth / resultWidth :: Double
--- a/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
+++ b/ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
@@ -4,6 +4,7 @@ module Ariadne.UI.Vty.Scrolling
      , handleScrollingEvent
      , scrollToEnd
      , fixedViewport
+     , viewportWithScrollBar
      ) where
 
 import Universum
@@ -11,7 +12,9 @@ import Universum
 import Ariadne.UI.Vty.Keyboard
 
 import qualified Brick as B
+import qualified Brick.Widgets.Border as BR
 import qualified Graphics.Vty as V
+import qualified Control.Monad.Trans.Reader as R
 
 data ScrollingAction
   = ScrollingLineUp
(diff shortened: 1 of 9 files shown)


goodsoft:
Also, please avoid "updating branch", as it somewhat clutters PRs.
We usually rebase branches like this:
```
git fetch origin master:master
git rebase master
git push -f
```

pasqu4le:
Ok, will avoid the "update branch" button from now on.

Regarding the first 2 points, now that I know what `fixedViewport`is about I agree that the scrollbar should be extended to the widgets that use it. If you think that's a good idea I can try to merge the 2 functions in one. Also, the approach of modifying the context and calling `viewport` on it is what I was trying to do in the beginning, but I don't know how to retrieve the sizes of the widget "inside" the viewport without rendering this one before. I needed those functions from brick to do this. Anyway, I don't like to copy those as well, so I'll gladly look into this tomorrow. 

For your third point, I did notice that and I confirm that some widgets look better when the scrollbar take the place of the right padding. I ended up avoiding that because of course for other widgets it would have cut a column. But yes, getting rid of that padding it would solve this. Will do.

goodsoft:
You can continue working in this PR, just commiting into this branch, no need to close it.

pasqu4le:
I'll be honest and just tell you that I never meant to close it, but my clumsy finger on a smartphone did.

goodsoft:
> If you think that's a good idea I can try to merge the 2 functions in one.
I think we should carefully consider all viewport usages, because some of them might not need a scrollbar.
For example, `Edit` widget was a horizontal viewport at some point to allow typing past right edge.

> Also, the approach of modifying the context and calling viewport on it is what I was trying to do in the beginning, but I don't know how to retrieve the sizes of the widget "inside" the viewport without rendering this one before.
It basically goes like this:
1. get current rendering context with currently available width/height
2. adjust context accordingly for inner widget
3. render the inner widget with adjusted context
4. fix up the resulting `V.Image`
You basically do that already, just that most of the code can be easily "placed back into" `viewport`

pasqu4le:
Ok, so I pushed a little commit that as you suggested pushes things "back into" `viewport`.

> I think we should carefully consider all viewport usages, because some of them might not need a scrollbar.
> For example, `Edit` widget was a horizontal viewport at some point to allow typing past right edge.

You are right. It seems that the wigets using `fixedViewport`s are:
- `Edit` that does not need a scrollbar
- `Logs` that seems to be working fine with the current version of the scrollbar (except for a small fix that I will do soon)
- `Tree` that would work fine with this scrollbar except that as it is right now it behaves greedily and splits the opening view in half.

So `Logs` does not have a problem, `Tree` would work if the 2 functions get merged, but `Edit` would still need a `fixedViewport`. I'm not sure it's a good idea to merge them after all.

Anyway, I'll now go fixing that little bug and remove the aesthetic padding on those widgets.

pasqu4le:
Update: the `Repl` was the only one with a "real" offset.
`Wallet`, `AddWallet` and `Account` all have a right single-char padding too, but inside the `viewport`.

So there are 2 options for the `Repl`:
- move the padding inside the `viewport`
- just remove the right-padding (that I did in this commit)

There are 2 options for the other ones too:
- leave the padding where it is (it's inside, so not a probem)
- leave the padding there, but remove the right one (that I did)

There is little difference between the options and the offset is gone anyway, but I thought to let you know.
Also, still have some problem with the `Logs` widget.

ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
```diff
@@ -84,3 +87,61 @@ fixedViewport name vpType p = case vpType of
...
+    -- add the scrollbar to the result
+    mvp <- B.unsafeLookupViewport name
+    let vpWidget = B.viewport name vpType $
+            B.Widget (B.hSize p) (B.vSize p) (return result)
+
+        resultImg = B.image result
+        (availH, availW) = case vpType of
+            B.Both -> (B.availHeight c - 1, B.availWidth c - 1)
+            _ -> (B.availHeight c, B.availWidth c) 
+
```
goodsoft: Please remove trailing spaces here

ui/vty-lib/src/Ariadne/UI/Vty/Scrolling.hs
```diff
@@ -84,3 +87,61 @@ fixedViewport name vpType p = case vpType of
...
+        enoughWidth = V.imageWidth resultImg <= availW
+        availWidth = fromIntegral availW
+        resultWidth = fromIntegral $ V.imageWidth resultImg
+        vpPadLeft = fromIntegral $ maybe 0 (view B.vpLeft) mvp
+        hRatio = availWidth / resultWidth :: Double
+        hPad = round $ hRatio * vpPadLeft
+        hSize = max 1 . round $ hRatio * availWidth
+        hBar = B.hLimit (hPad + hSize) $ B.padLeft (B.Pad hPad) BR.hBorder
+        addHBar widg = if enoughWidth then widg else B.vBox [widg, hBar]
```
goodsoft: I would create one function like `availDim -> imageDim -> (barSize, barPad)` and use it for both horizontal or vertical bars

pasqu4le:
I will fix the git problem first thing, I don't have uncommitted changes, so it shouldn't be too hard. As for the commit message, I forgot about the ID, sorry.

As for the changes, I will leave `Repl` as it is now and put the 1-char padding back in `AddWallet`/`Wallet`/`Account`, it will take but a moment.

> The Logs widget might be problematic to implement scrollbar for, because it's not always actually scrollable. I think we might just make a custom indicator (like, v symbol in the bottom-right corner), when it's not in the "scroll follows content" mode, i.e. when you scrolled it up.

Yeah I figured the problem was something like that because it was kinda working, but behaved differently when on the bottom. Anyway, this seems like a problem of his own, should the `v`be made directly in the `Logs` code? Or should I make another function just to do this?

> Also, I think that you should integrate the constraint releasing logic into fixedViewport, and then use fixedViewport in viewportWithScrollBar (maybe rename it into something shorter, like scrollingViewport).
Then you will use fixedViewport for Edit, and scrollingViewport for Tree.
I consider the fixedViewport logic more correct than B.viewport with regards to child sizing policy, and the only thing stopping me from using it everywhere was not releasing children constraints.
Now that you have it implemented, we can use it for every viewport out there.

Seem the best approach to me too, I'll implement this as well.

goodsoft:
> Anyway, this seems like a problem of his own, should the `v` be made directly in the Logs code?

Yeah, as the rest of custom scrolling logic is there anyway

pasqu4le:
Good, will do that as well then

pasqu4le:
I did `git reset --hard origin/pasqu4le/ad262-viewport-scrollbar` and then `git rebase master`, but now I cannot build anymore, because it `Could not find module ‘IiExtras’`. What should I do?

pasqu4le:
I think I made all the necessary modifications. I'm not particularly satisfied with the `Logs` indicator, but nothing better came to my mind.

goodsoft:
Oh, I now remember, why I wasn't releasing constraints in `fixedViewports`.
That way `Edit` widget had a context with adequate `availWidth`.
But now it is 100000 characters wide and inputs don't wrap as I intended :(

pasqu4le:
Oh ok, if this is the case I should just let it render the inner widget without releasing.
Anything else?

goodsoft:
Everything else looks good :+1:

PR #292, opened 2018-08-28
[AD-331] Add dialog to remove wallet

maksbotan:
**YT issue:** https://issues.serokell.io/issue/AD-331

**Checklist:**

  - [x] [README](README.md)

**Description:**

### DIFF
--- a/ui/qt-app/Glue.hs
+++ b/ui/qt-app/Glue.hs
@@ -136,6 +136,9 @@ knitFaceToUI UiFace{..} KnitFace{..} =
       UiNewAddress -> do
         Right $ Knit.ExprProcCall
           (Knit.ProcCall Knit.newAddressCommandName [])
+      UiRemoveCurrentItem -> do
+        Right $ Knit.ExprProcCall
+          (Knit.ProcCall Knit.removeCommandName [])
 
     resultToUI result = \case
       UiSend _ _ ->
--- a/ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
+++ b/ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
@@ -1,17 +1,20 @@
 module Ariadne.UI.Qt.Widgets.Dialogs.Delete
-  ( DeletingEntity(..)
-  , DeletionResult
+  ( DeletingItem(..)
+  , DeletionResult(..)
   , runDelete
   ) where
 
 import Universum
 
 import qualified Data.Text as T
+import Formatting
 
 import Graphics.UI.Qtah.Signal (connect_)
+import Graphics.UI.Qtah.Widgets.QSizePolicy (QSizePolicyPolicy(..))
 
 import qualified Graphics.UI.Qtah.Widgets.QAbstractButton as QAbstractButton
 import qualified Graphics.UI.Qtah.Widgets.QBoxLayout as QBoxLayout
+import qualified Graphics.UI.Qtah.Widgets.QHBoxLayout as QHBoxLayout
 import qualified Graphics.UI.Qtah.Widgets.QCheckBox as QCheckBox
 import qualified Graphics.UI.Qtah.Widgets.QDialog as QDialog
 import qualified Graphics.UI.Qtah.Widgets.QLabel as QLabel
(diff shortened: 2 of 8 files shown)


ui/qt-app/Glue.hs
```diff
@@ -136,6 +136,9 @@ knitFaceToUI UiFace{..} KnitFace{..} =
       UiNewAddress -> do
         Right $ Knit.ExprProcCall
           (Knit.ProcCall Knit.newAddressCommandName [])
+      UiRemoveCurrentItem -> do
```
maksbotan: Should this command also have a corresponding result? I don't think so, if only to show error.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
```diff
@@ -0,0 +1,134 @@
...
+     \Otherwise you will lose all your funds connected to this " % itemTypeFormat % ".")
+    itemType
+  QLabel.setWordWrap isSureLabel True
+  QWidget.setMinimumSizeRaw isSureLabel 600 30
+  isSureLayout <- QHBoxLayout.new
+  QBoxLayout.addWidget isSureLayout isSure
+  QBoxLayout.addWidget isSureLayout isSureLabel
+  QBoxLayout.addLayout layout isSureLayout
+
+  -- TODO handle empty name case
```
maksbotan: See question re `Maybe`. If we were to allow empty names, we'd need to disable retype field entirely in those cases, I guess.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/NewWallet.hs
```diff
@@ -163,16 +117,13 @@ initNewWallet = do
...
-  QLayout.setContentsMarginsRaw layout 24 24 24 24
+  layout <- createLayout newWallet
 
   modeSelector <- createModeSelector
 
-  QBoxLayout.addWidget layout modeSelector
+  addHeader layout modeSelector
   void $ QLayout.setWidgetAlignment layout modeSelector $ alignHCenter .|. alignVCenter
-  QBoxLayout.addSpacing layout 42
+  void $ setProperty modeSelector ("styleRole" :: Text) ("dialogHeader" :: Text)
```
maksbotan: I changed this 42 spacing into 24 (in `Util.hs`). All other dialogs in designs have 24 px spacing, I believe 42px in new wallet is an error or just some oddity.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
```diff
@@ -180,10 +194,15 @@ handleWalletInfoEvent UiLangFace{..} ev = do
               UiSelectionWallet UiWalletInfo{..} -> (uwiLabel, uwiBalance, True)
               UiSelectionAccount UiAccountInfo{..} -> (uaciLabel, uaciBalance, False)
 
-      whenJust itemName $ QLabel.setText itemNameLabel . toString . toUpper
+      QLabel.setText itemNameLabel . toString . toUpper . fromMaybe "" $ itemName
```
maksbotan: I believe this `Maybe` is pointless and misleading. How can wallet / account not have a name? I'd rather drop it.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
```diff
@@ -53,6 +54,9 @@ data WalletInfo =
     , itemModel :: QStandardItemModel.QStandardItemModel
     , selectionModel :: QItemSelectionModel.QItemSelectionModel
     , createAccountButton :: QPushButton.QPushButton
+    , currentItemType :: IORef (Maybe DeletingItem)
```
maksbotan: Is it ok to use `DeletingItem` type here?

gromakovsky:
I think that for account deletion it's enough to ask whether the user really wants to delete it. Accounts can be restored as long as there is access to the root wallet. So I think asking to type the name is too much.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
```diff
@@ -0,0 +1,134 @@
...
+  connect_ deleteButton QAbstractButton.clickedSignal $ \_ -> QDialog.accept delete
+
+  revalidate del
+
+  return del
+
+runDelete :: DeletingItem -> Text -> IO DeletionResult
+runDelete itemType itemName = do
+  del@Delete{delete = delete} <- initDelete itemType itemName
+  result <- fmap toEnum . liftIO . QDialog.exec $ delete
```
gromakovsky: `liftIO` is redundant.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
gromakovsky: `Maybe` is a mistake it seems.

maksbotan: Okay, I'll drop it later.

ui/qt-app/Glue.hs
gromakovsky: > I don't think so

same

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
```diff
@@ -53,6 +54,9 @@ data WalletInfo =
...
+    , currentItemName :: IORef (Text)
```
gromakovsky: `()` are redundant, btw.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
gromakovsky: I don't like it, defining a new data type seems better, because set of item types which can be deleted doesn't necessary have to be the same as set of item types which can be selected.

PR #297, opened 2018-09-01
[AD-334] Mnemonic confirmation dialog

maksbotan:
**YT issue:** https://issues.serokell.io/issue/AD-334

**Checklist:**

  - [x] [README](README.md)

**Description:**
A dialog to confirm mnemonic. Some questions are present.

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -19,9 +19,13 @@ module Ariadne.Wallet.Backend.KeyStorage
          -- * Exceptions
        , NoWalletSelection (..)
        , NoAccountSelection (..)
+
+         -- * Util
+       , generateMnemonic
        ) where
 
 import Universum
+import qualified Universum.Unsafe as Unsafe
 
 import Control.Exception (Exception(displayException))
 import Control.Lens (ix, (%=))
@@ -339,9 +343,9 @@ newAccount acidDb WalletFace{..} walletSelRef walletRef mbAccountName = do
   walletDb <- query acidDb Snapshot
   (rootId,accList)  <- second toList <$>
         resolveWalletRefThenRead walletSelRef walletRef walletDb readAccountsByRootId
-  
+
   let namesVec = V.fromList (map (unAccountName . (^. hdAccountName)) accList)
-  
+
   accountName <- case (unAccountName <$> mbAccountName) of
     Nothing ->
       return (mkUntitled "Untitled account " namesVec)
@@ -373,6 +377,13 @@ instance Buildable InvalidEntropySize where
 instance Exception InvalidEntropySize where
     displayException = toString . pretty
 
+generateMnemonic :: Byte -> IO [Text]
+generateMnemonic entropySize = do
+  unless (entropySize `elem` [16, 20, 24, 28, 32]) $
+      throwM $ InvalidEntropySize entropySize
+  mnemonic <- entropyToMnemonic <$> secureRandomBS (fromIntegral entropySize)
+  return $ mnemonic ++ ["ariadne-v0"]
+
 -- | Generate a mnemonic and a wallet from this mnemonic and add the
 -- wallet to the storage.
 newWallet ::
@@ -386,17 +397,15 @@ newWallet ::
     -> IO [Text]
 newWallet acidDb walletConfig face runCardanoMode pp mbWalletName mbEntropySize = do
   let entropySize = fromMaybe (wcEntropySize walletConfig) mbEntropySize
-  unless (entropySize `elem` [16, 20, 24, 28, 32]) $
-      throwM $ InvalidEntropySize entropySize
-  entropy <- secureRandomBS (fromIntegral entropySize)
-  let mnemonic = entropyToMnemonic entropy
+  mnemonic <- generateMnemonic entropySize
   -- The empty string below is called a passphrase in BIP-39, but
   -- it's essentially an extra mnemonic word. We consider it an
   -- advanced feature and do not provide it for now.
-  let seed = mnemonicToSeed (unwords mnemonic) ""
+  -- generateMnemonic function guarantees that mnemonic is a non-empty list
+  let seed = mnemonicToSeed (unwords $ Unsafe.init mnemonic) ""
   let (_, esk) = safeDeterministicKeyGen seed pp
   hasPass <- mkHasPass runCardanoMode (pp == emptyPassphrase)
-  mnemonic ++ ["ariadne-v0"] <$
+  mnemonic <$
     addWallet acidDb face runCardanoMode esk mbWalletName mempty hasPass assurance
   where
     -- TODO(AD-251): allow selecting assurance.
--- a/ui/qt-app/Glue.hs
+++ b/ui/qt-app/Glue.hs
@@ -111,13 +111,6 @@ knitFaceToUI UiFace{..} KnitFace{..} =
                 ]
             ]
           )
-      UiNewWallet name password -> do
-        Right $ Knit.ExprProcCall
-          (Knit.ProcCall Knit.newWalletCommandName $
-            [Knit.ArgKw "name" . Knit.ExprLit . Knit.toLit . Knit.LitString $ name]
-            ++ maybe []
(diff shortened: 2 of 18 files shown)


ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/ConfirmMnemonic.hs
```diff
@@ -0,0 +1,233 @@
...
+  QBoxLayout.addWidget layout instructionLabel
+
+  (noOneLooksWidget, noOneLooksLayout) <- createSubWidget
+  noOneLooks <- createCheckBox noOneLooksLayout
+    "Make sure nobody looks into your screen unless you want them to have access to your funds."
+  QBoxLayout.addWidget layout noOneLooksWidget
+
+  mnemonicWidget <- QLabel.new
+  void $ setProperty mnemonicWidget ("styleRole" :: Text) ("mnemonicDisplay" :: Text)
+  QLabel.setTextInteractionFlags mnemonicWidget TextSelectableByMouse
```
maksbotan: I believe allowing selecting (and therefore copying) is a good think to do, right?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
```diff
@@ -102,8 +102,7 @@ handleWalletTreeEvent UiLangFace{..} ev = do
   lift $ case ev of
     WalletTreeNewWalletCommandResult _commandId result -> case result of
       UiNewWalletCommandSuccess mnemonic -> do
-        void $ QMessageBox.information treeView ("Success" :: String) $
-          toString $ "This is your wallet mnemonic. Save it.\n\n" <> intercalate " " mnemonic
+        void $ runConfirmMnemonic mnemonic
```
maksbotan: If user fails this dialog, the wallet is still created. In fact, it is already created by the time this dialog starts. Do we want to delete it if user can't confirm he wrote down the mnemonic? I'm not sure what UX Alexander meant with this.

gromakovsky: User should see that the wallet has been created only after they have entered the correct recovery phrase. And that's a problem because currently mnemonic generation and wallet creation are coupled together. We need to decouple them.
I suppose we need a new command which generates a mnemonic and returns it, without creating a wallet. And then we can call `restore` with this mnemonic and `full:false` when the mnemonic is entered correctly.

ui/qt-app/Glue.hs
```diff
@@ -331,3 +322,9 @@ historyToUI ch = UiHistoryFace
   , historyPrevCommand = toPrevCommand ch
   }
 
+----------------------------------------------------------------------------
```
maksbotan: This does not fit into any of our current faces. In face, we only have `langFace` exposed to the UI and I don't think `generateMnemonic` fits there.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
```diff
@@ -81,17 +83,27 @@ initWalletTree langFace itemModel selectionModel = do
...
       case nwSpecifier of
-        NewWalletName -> UiNewWallet nwName nwPassword
-        NewWalletMnemonic mnemonic full -> UiRestoreWallet nwName nwPassword mnemonic full
+        NewWalletName -> doCreateNewWallet langFace walletTree nwName nwPassword
+        NewWalletMnemonic mnemonic full -> void $ langPutUiCommand $
+          UiRestoreWallet nwName nwPassword mnemonic full
+
+doCreateNewWallet :: UiLangFace -> WalletTree -> Text -> Maybe Text -> IO ()
+doCreateNewWallet UiLangFace{..} WalletTree{..} name password = do
+  let entropySize = 16 -- FIXME Hardcoded for now, will have to fetch from wallet config
```
maksbotan: For that we need to expose wallet configuration to the ui. At this step I could've hardcoded entropy size from the config into `generateMnemonic` in `Main.hs` (or can I?), but in the future I believe we want to preselect default option in the dialog, so configuration should be exposed anyway.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
```diff
@@ -81,17 +83,27 @@ initWalletTree langFace itemModel selectionModel = do
...
+  mnemonic <- mnemonicGenerator entropySize
+  confirmationResult <- runConfirmMnemonic mnemonic
+
+  case confirmationResult of
+    ConfirmMnemonicSuccess -> void $ langPutUiCommand $ UiRestoreWallet name password (unwords mnemonic) False
+    ConfirmMnemonicFailure -> return ()
```
maksbotan: Do we want some message box in this case?

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -388,15 +394,14 @@ newWallet acidDb walletConfig face runCardanoMode pp mbWalletName mbEntropySize
   let entropySize = fromMaybe (wcEntropySize walletConfig) mbEntropySize
   unless (entropySize `elem` [16, 20, 24, 28, 32]) $
       throwM $ InvalidEntropySize entropySize
-  entropy <- secureRandomBS (fromIntegral entropySize)
-  let mnemonic = entropyToMnemonic entropy
+  mnemonic <- generateMnemonic entropySize
   -- The empty string below is called a passphrase in BIP-39, but
   -- it's essentially an extra mnemonic word. We consider it an
   -- advanced feature and do not provide it for now.
   let seed = mnemonicToSeed (unwords mnemonic) ""
```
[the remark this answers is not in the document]
maksbotan: Hmmm... This means we need to either inline `generateMnemonic` back or append `ariadne-v0` in the GUI separately. Or just take `init` here. I'm not sure which is better. The last one probably.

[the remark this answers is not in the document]
gromakovsky: > I'm not sure which is better. The last one probably.

:same:

Also I think that entropy size check should be part of `generateMnemonic`.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
gromakovsky: Can you ask in #ariadne-design please?

ui/qt-app/Glue.hs
gromakovsky: I think it should be a separate function (as it is), but defined inside the `Glue between the Wallet backend and Qt frontend` block (and exported right after `putWalletEventToUI`).

ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
```diff
@@ -210,3 +205,5 @@ data UiAddressInfo = UiAddressInfo
 data UiSelectionInfo
   = UiSelectionWallet !UiWalletInfo
   | UiSelectionAccount !UiAccountInfo
+
+type MnemonicGenerator = Int -> IO [Text]
```
gromakovsky: AFAIU, this module exports various stuff related to *what UI can do*, not *what UI needs*. So I think this alias should be somewhere in another place (maybe in `WalletTree.hs`).

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
gromakovsky: I suppose we need to pass not only `MnemonicGenerator`, but also `entropySize :: Byte` or `:: IO Byte` to UI creation function. `:: Byte` should be enough now, `IO Byte` will be needed if we make it possible to update configuration in runtime.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletTree.hs
maksbotan: Asked.

maksbotan:
This feels like a monstrosity now :man_shrugging:

PR #298, opened 2018-09-03
[AD-64] Hide passwords in Knit history

pasqu4le:
**YT issue:** https://issues.serokell.io/issue/AD-64

**Checklist:**

  - [ ] [README](README.md)

**Description:**

The password fields are now hidden as `*`s by the `ppExpr` pretty-printing function and stored like this in the history as well.

### DIFF
--- a/ui/vty-app/Glue.hs
+++ b/ui/vty-app/Glue.hs
@@ -65,7 +65,7 @@ knitFaceToUI
   -> UiLangFace
 knitFaceToUI UiFace{..} KnitFace{..} =
   UiLangFace
-    { langPutCommand = putCommand commandHandle
+    { langPutCommand = putCommand Nothing
     , langPutUiCommand = putUiCommand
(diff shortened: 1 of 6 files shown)


knit/src/Knit/Printer.hs
```diff
@@ -75,8 +75,14 @@ ppExpr =
 
     ppArg = \case
       ArgPos a -> parensIfProcCall a (ppExpr a)
-      ArgKw name a -> nameToDoc name PP.<> PP.colon PP.<+>
-        parensIfProcCall a (ppExpr a)
+      ArgKw name a -> nameToDoc name PP.<> PP.colon PP.<+> (
```
int-index: That's a very bad place to add this logic. What if the password is passed as a positional argument?

pasqu4le: You are right, I will.

PR #299, opened 2018-09-04
[AD-270] Check keyfile for unknown keys on startup

Haskell-mouse:
**YT issue:** https://issues.serokell.io/issue/AD-270

**Checklist:**

  - [ ] [README](README.md)

**Description:**

> It's possible that key file has more secret keys than there are wallets in wallet DB. Btw, the opposite is impossible. 

So - now unknown keys are deleted on startup.

### DIFF
--- a/ariadne/cardano/src/Ariadne/MainTemplate.hs
+++ b/ariadne/cardano/src/Ariadne/MainTemplate.hs
@@ -7,8 +7,8 @@ module Ariadne.MainTemplate
 
 import Control.Concurrent (forkIO, myThreadId, throwTo)
 import Control.Concurrent.Async
-  (Async(..), AsyncCancelled(..), ExceptionInLinkedThread(..), race_,
-  waitCatch, withAsync)
+  (Async(..), AsyncCancelled(..), ExceptionInLinkedThread(..), concurrently,
+  race_, waitCatch, withAsync)
 import Control.Monad.Component (ComponentM, runComponentM)
 import Control.Natural (($$))
 import Data.Version (Version)
@@ -133,7 +133,8 @@ initializeEverything MainSettings {..}
   let
     mkWalletFace :: (Doc -> IO ()) -> WalletFace
     walletInitAction :: IO ()
-    (mkWalletFace, walletInitAction) = mkWallet
+    postInitAction :: IO ()
+    (mkWalletFace, walletInitAction, postInitAction) = mkWallet
 
     knitExecContext ::
         (Doc -> IO ()) -> Knit.ExecContext IO (AllComponents uiComponents)
@@ -167,24 +168,24 @@ initializeEverything MainSettings {..}
     initAction :: IO ()
     initAction = walletInitAction
 
+    postStartupAction :: IO ()
+    postStartupAction = postInitAction
+
     serviceAction :: IO ()
     serviceAction =
-      raceWithUpdateCheckAction $
-      cardanoAction
+      raceWithUpdateCheckAction cardanoAction
 
     mainAction :: IO ()
     mainAction = do
       initAction
-
       -- Spawn backend actions in async thread, then run ui action in the main thread
       -- This is needed because some UI libraries (Qt) insist on livng in the main thread
-      withAsync serviceAction $ \serviceThread -> do
-        -- Make custom link to service thread for UI apps.
-        -- It is going to call msPutBackendErrorToUI and rethrow exception, if something goes wrong
-        linkUI serviceThread $ msPutBackendErrorToUI uiFace
-        logDebug logging "Launching the UI..."
-        uiAction
-
+      withAsync (concurrently postStartupAction serviceAction) $ \serviceThread -> do
+          -- Make custom link to service thread for UI apps.
+          -- It is going to call msPutBackendErrorToUI and rethrow exception, if something goes wrong
+          linkUI serviceThread $ msPutBackendErrorToUI uiFace
+          logDebug logging "Launching the UI..."
+          uiAction
   return mainAction
 
 -- Similar to link from Control.Concurrent.Async, but calls finalizer on exception
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend.hs
@@ -30,7 +30,7 @@ import Ariadne.Wallet.Face
 -- | This is what we create initially, before actually creating 'WalletFace'.
 data WalletPreface = WalletPreface
     { wpBListener :: !BListenerHandle
-    , wpMakeWallet :: !((Doc -> IO ()) -> WalletFace, IO ())
+    , wpMakeWallet :: !((Doc -> IO ()) -> WalletFace, IO (), IO ())
     }
 
 createWalletBackend
@@ -92,7 +92,7 @@ createWalletBackend walletConfig cardanoFace sendWalletEvent getPass voidPass lo
             sendWalletEvent $ WalletRequireConfirm mVar confirmType
             takeMVar mVar
 
-        mkWallet = (mkWalletFace, initWalletAction)
+        mkWallet = (mkWalletFace, initWalletAction, postInitAction)
           where
             NT runCardanoMode = cardanoRunCardanoMode cardanoFace
             withDicts :: ((HasConfigurations, HasCompileInfo) => r) -> r
@@ -119,8 +119,11 @@ createWalletBackend walletConfig cardanoFace sendWalletEvent getPass voidPass lo
                 , walletBalance = getBalance pwl walletSelRef
                 , walletSumCoins = \amounts -> return $ unsafeIntegerToCoin $ sumCoins amounts
                 }
-            initWalletAction =
-                refreshState pwl walletSelRef sendWalletEvent
+            initWalletAction = refreshState pwl walletSelRef sendWalletEvent
+
+            postInitAction = do
+                checkUnknownKeys pwl waitUiConfirm
+                checkWalletsWithoutSecretKey pwl waitUiConfirm
 
         walletPreface = WalletPreface
             { wpBListener = bListenerHandle
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -13,6 +13,8 @@ module Ariadne.Wallet.Backend.KeyStorage
        , getBalance
        , renameSelection
        , removeSelection
+       , checkUnknownKeys
+       , checkWalletsWithoutSecretKey
        , deriveBip44KeyPair
 
          -- * Exceptions
@@ -26,8 +28,10 @@ import qualified Universum.Unsafe as Unsafe (init)
 
 import Control.Exception (Exception(displayException))
 import Control.Lens (ix)
+
+import qualified Data.Text as T
 import qualified Data.Text.Buildable
-import Formatting (bprint, int, (%))
+import Formatting (bprint, int, sformat, (%))
 import Serokell.Data.Memory.Units (Byte)
 
 import Pos.Core (mkCoin)
@@ -425,6 +429,31 @@ renameSelection pwl WalletFace{..} walletSelRef name = do
         throwLeftIO $ void <$> pwlUpdateAccountName pwl hdAccId (AccountName name)
 
   walletRefreshState
+
+-- | This function removes unknown keys from key file on startup, if user confirms that.
+checkUnknownKeys
+  :: PassiveWalletLayer IO
+  -> (ConfirmationType -> IO Bool)  -- ^ Asks user permission to delete unknown keys
+  -> IO ()
+checkUnknownKeys pwl confirmation = do
+  unknownRootIds <- pwlGetUnknownKeys pwl
+  unless (null unknownRootIds) $ do
+    let unknownRootIdsText =
+          T.dropEnd 1 . unlines $ sformat hashHexF . _fromDb . getHdRootId <$> unknownRootIds
+    whenM (confirmation $ ConfirmDelUnknownKeys unknownRootIdsText) $
+      pwlRemoveUnknownKeys pwl unknownRootIds
+
+-- | This function removes wallets with missed secret keys, if user confirms that.
+checkWalletsWithoutSecretKey
+  :: PassiveWalletLayer IO
+  -> (ConfirmationType -> IO Bool)  -- ^ Asks user permission to delete such wallets
+  -> IO ()
+checkWalletsWithoutSecretKey pwl confirmation = do
+  (walletNamesWithMissedSecretKeys, rootIDsWithMissedSecretKeys) <- pwlGetWalletsWithoutSecretKeys pwl
+  let walletsWithoutKeysText = unlines $ unWalletName <$> walletNamesWithMissedSecretKeys
+  unless (null rootIDsWithMissedSecretKeys) $ do
+    whenM (confirmation $ ConfirmDelBrokenWallets $ T.dropEnd 1 $ walletsWithoutKeysText) $
+      mapM_ (pwlDeleteWallet pwl) rootIDsWithMissedSecretKeys
 
 {-------------------------------------------------------------------------------
   Utilities
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
@@ -1,0 +1,62 @@
+-- | This module contains functions that are needed
+-- to check consistency of wallets and key storage.
+
+module Ariadne.Wallet.Cardano.Kernel.Consistency
+       ( getUnknownKeys
+       , getUnknownWallets
+       ) where
+
+import Data.List ((\\))
+import qualified Data.Map as Map
+import Fmt (blockListF, (+|), (|+))
+
+import Ariadne.Wallet.Cardano.Kernel (PassiveWallet, getWalletSnapshot)
+import Ariadne.Wallet.Cardano.Kernel.DB.AcidState (dbHdWallets)
+import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet
+  (HdRoot, HdRootId, WalletName, hdRootId, hdRootName)
+import Ariadne.Wallet.Cardano.Kernel.Keystore (Keystore)
+import Ariadne.Wallet.Cardano.Kernel.Types (WalletId(..))
+
+import qualified Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Read as HDRead
+import qualified Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet as IxSet
+import qualified Ariadne.Wallet.Cardano.Kernel.Keystore as Keystore
+
+-- | Checks keystore for unknown keys and return list of them.
+getUnknownKeys :: MonadIO m => PassiveWallet -> Keystore -> m [HdRootId]
+getUnknownKeys pwallet keystore = do
+  (keystoreRootIds,walletRootIds,_) <- getRootIdsFromWalletAndKeyfile pwallet keystore
+  return $ (keystoreRootIds \\ walletRootIds)
+
+-- | Returns list of wallets without corresponding key in keystore.
+getUnknownWallets
+  :: MonadIO m => PassiveWallet -> Keystore -> m ([WalletName], [HdRootId])
+getUnknownWallets pwallet keystore = do
+  (keystoreRootIds,walletRootIds, wallets) <- getRootIdsFromWalletAndKeyfile pwallet keystore
+  let rootIDsWithMissedSecretKeys = walletRootIds \\ keystoreRootIds
+  let eitherWalletsWithMissedSecretKeys =
+        mapM (flip IxSet.getTheOnly wallets) rootIDsWithMissedSecretKeys
+  let walletNamesWithMissedSecretKeys = case eitherWalletsWithMissedSecretKeys of
+        Right wallet -> view hdRootName <$> wallet
+        Left  (brokenRootId, brokenWallets) ->
+                 error $ "There are more (or less) than one wallet with the same HdRootId: \n \
+                         \Broken RootId: " +|brokenRootId|+ "\n" <>
+                         "Names of wallets with this id: " <>
+                         blockListF (view hdRootName <$> IxSet.toList brokenWallets)
+  return (walletNamesWithMissedSecretKeys,rootIDsWithMissedSecretKeys)
+
+-- | helper function. It returns list of HdRootIds from keystore and from wallets database.
+-- And also ixset of all HdRoots
+getRootIdsFromWalletAndKeyfile
+  :: MonadIO m
+  => PassiveWallet
+  -> Keystore
+  ->  m ([HdRootId], [HdRootId], IxSet.IxSet HdRoot)
+getRootIdsFromWalletAndKeyfile pwallet keystore = do
+  snapshot <- liftIO $ getWalletSnapshot pwallet
+  let wallets = HDRead.readAllHdRoots (snapshot ^. dbHdWallets)
+  let walletRootIds = view hdRootId <$> IxSet.toList wallets
+  (keystoreRootIds) <- liftIO $ Map.keys <$> Keystore.toMap keystore
+  return (getHdRootId <$> keystoreRootIds, walletRootIds, wallets)
+  where
+    getHdRootId :: WalletId -> HdRootId
+    getHdRootId (WalletIdHdSeq rootId) = rootId
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
@@ -18,6 +18,7 @@ module Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet
        , member
        , size
        , getOne
+       , getTheOnly
        , toMap
          -- * Modification
        , updateIxManyM
@@ -41,7 +42,7 @@ import Data.Coerce (coerce)
 import qualified Data.Foldable
 import qualified Data.IxSet.Typed as IxSet
 import qualified Data.Map.Strict as Map
-import Data.SafeCopy (SafeCopy(..))
+import Data.SafeCopy (SafeCopy(..), contain, safeGet, safePut)
 import qualified Data.Set as Set
 import qualified Data.Traversable
 
@@ -84,6 +85,10 @@ instance HasPrimKey a => Ord (OrdByPrimKey a) where
 instance Buildable a => Buildable (OrdByPrimKey a) where
     build (WrapOrdByPrimKey o) = bprint build o
 
+instance (SafeCopy a) => SafeCopy (OrdByPrimKey a) where
+  getCopy = contain $ WrapOrdByPrimKey <$> safeGet
+  putCopy (WrapOrdByPrimKey a) = contain . safePut $ a
+
 {-------------------------------------------------------------------------------
   Wrap IxSet
 -------------------------------------------------------------------------------}
@@ -116,9 +121,9 @@ type IsIndexOf ix a = IxSet.IsIndexOf ix (PrimKey a ': IndicesOf a)
   Safecopy
 -------------------------------------------------------------------------------}
 
-instance SafeCopy a => SafeCopy (IxSet a) where
-  getCopy = error "getCopy for IxSet wrapper"
-  putCopy = error "putCopy for IxSet wrapper"
+instance (Indexable a, SafeCopy a) => SafeCopy (IxSet a) where
+  getCopy =  contain $ WrapIxSet <$> safeGet
+  putCopy (WrapIxSet ixset) = contain . safePut $ ixset
 
 {-------------------------------------------------------------------------------
   Building 'Indexable' instances
@@ -174,6 +179,16 @@ instance Foldable IxSet where
 
 getEQ :: (Indexable a, IsIndexOf ix a) => ix -> IxSet a -> IxSet a
 getEQ ix = WrapIxSet . IxSet.getEQ ix . unwrapIxSet
+
+-- | Returns either element by its index or
+-- index to which several (or zero) elements correspond + those elements
+getTheOnly
+    :: (Indexable a, HasPrimKey a, IsIndexOf ix a) => ix -> IxSet a -> Either (ix, IxSet a) a
+getTheOnly ix ixset =
+    let eqElems = getEQ ix ixset
+    in case getOne eqElems of
+         Just a -> Right a
+         Nothing -> Left (ix, eqElems)
 
 member :: (HasPrimKey a, Indexable a) => PrimKey a -> IxSet a -> Bool
 member pk = isJust . view (Lens.at pk)
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
@@ -17,6 +17,7 @@ module Ariadne.Wallet.Cardano.Kernel.Keystore
        , lookup
        -- * Conversions
        , toList
+       , toMap
        -- * Tests handy functions
        , bracketTestKeystore
        ) where
@@ -174,12 +175,15 @@ delete walletId (Keystore mstorage fp) = do
 
 -- | Returns all the 'EncryptedSecretKey' known to this 'Keystore'.
 toList :: Keystore -> IO [(WalletId, EncryptedSecretKey)]
-toList (Keystore mstorage _) =
+toList keystore = Map.toList <$> toMap keystore
+
+toMap :: Keystore -> IO (Map WalletId EncryptedSecretKey)
+toMap (Keystore mstorage _) =
     withMVar mstorage $ \istorage ->
-        pure $ map (first hashPubKeyToWalletId) $ toPairs $ istorage ^. isWallets
-  where
-    hashPubKeyToWalletId :: AddressHash PublicKey -> WalletId
-    hashPubKeyToWalletId = WalletIdHdSeq . HdRootId . InDb
+      pure $ Map.mapKeys hashPubKeyToWalletId $ istorage ^. isWallets
+
+hashPubKeyToWalletId :: AddressHash PublicKey -> WalletId
+hashPubKeyToWalletId = WalletIdHdSeq . HdRootId . InDb
 
 {-------------------------------------------------------------------------------
   Utilities
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
@@ -3,6 +3,7 @@ module Ariadne.Wallet.Cardano.Kernel.Wallets
        , updateHdWalletName
        , updateHdWalletAssurance
        , deleteHdWallet
+       , removeKeysFromKeystore
        , HasNonemptyPassphrase(..)
        , mkHasPP
        , CreateWithAddress(..)
@@ -189,6 +190,13 @@ deleteHdWallet pw rootId = do
             -- STEP 2: Purge the key from the keystore.
             Keystore.delete (WalletIdHdSeq rootId) (pw ^. walletKeystore)
             return $ Right ()
+
+removeKeysFromKeystore :: PassiveWallet
+                       -> [HD.HdRootId]
+                       -> IO ()
+removeKeysFromKeystore pw rootIds =
+  let keystore = pw ^. walletKeystore
+  in mapM_ (\rootId -> Keystore.delete (WalletIdHdSeq rootId) keystore) rootIds
 
 {-------------------------------------------------------------------------------
   Wallet update
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Kernel.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Kernel.hs
@@ -22,13 +22,17 @@ import qualified Ariadne.Wallet.Cardano.Kernel.Actions as Actions
 import qualified Ariadne.Wallet.Cardano.Kernel.Addresses as Kernel
 import qualified Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Read as HDRead
 import qualified Ariadne.Wallet.Cardano.Kernel.Keystore as Keystore (lookup)
+
 import qualified Ariadne.Wallet.Cardano.Kernel.Restore as Kernel
+
 import qualified Ariadne.Wallet.Cardano.Kernel.Wallets as Kernel
 
 import Ariadne.Wallet.Cardano.Kernel.Bip39 (mnemonicToSeedNoPassword)
+import Ariadne.Wallet.Cardano.Kernel.Consistency
 import Ariadne.Wallet.Cardano.Kernel.DB.AcidState (DB, dbHdWallets, defDB)
 import Ariadne.Wallet.Cardano.Kernel.DB.Resolved (ResolvedBlock)
 import Ariadne.Wallet.Cardano.Kernel.Keystore (Keystore)
+
 import Ariadne.Wallet.Cardano.Kernel.Types
   (AccountId(..), RawResolvedBlock(..), WalletId(..), fromRawResolvedBlock)
 import Ariadne.Wallet.Cardano.Kernel.Wallets (CreateWithAddress(..))
@@ -163,6 +167,11 @@ passiveWalletLayerCustomDBComponent logFunction keystore acidDB pm = do
             , pwlGetDBSnapshot         = liftIO $ Kernel.getWalletSnapshot wallet
             , pwlLookupKeystore        = \hdrId -> liftIO $
                 Keystore.lookup (WalletIdHdSeq hdrId) (wallet ^. Kernel.walletKeystore)
+            , pwlRemoveUnknownKeys     = \rootIds -> liftIO $
+                Kernel.removeKeysFromKeystore wallet rootIds
+            , pwlGetUnknownKeys        = getUnknownKeys wallet keystore
+            , pwlGetWalletsWithoutSecretKeys =  getUnknownWallets wallet keystore
+
             }
 
     -- The use of the unsafe constructor 'UnsafeRawResolvedBlock' is justified
@@ -195,3 +204,4 @@ activeWalletLayerComponent pwl pw = do
         , awlNewPending      = \hdAccId tx ->
             liftIO $ Kernel.newPending aw hdAccId tx
         }
+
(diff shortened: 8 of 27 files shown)


ariadne/cardano/src/Ariadne/Wallet/Backend.hs
```diff
@@ -30,7 +30,7 @@ import Ariadne.Wallet.Face
 data WalletPreface = WalletPreface
     { wpBListener :: !BListenerHandle
     , wpAddUserSecret :: !(TVar UserSecret -> IO ())
-    , wpMakeWallet :: !(CardanoFace -> ((Doc -> IO ()) -> WalletFace, IO ()))
+    , wpMakeWallet :: !(CardanoFace -> ((Doc -> IO ()) -> WalletFace, FilePath -> IO ()))
```
gromakovsky: Meh, it will be unnecessary after #288 because key file path will be in the Wallet configuration.

gromakovsky:
I suddenly realized that this solution is dangerous. If user somehow passes wrong path to wallet-db, all secret keys will be deleted and it may cause loss of funds. I see two options:
1. Use some heuristics to decide whether to delete secret keys (e. g. if set of keys is _similar_ to set of wallets).
2. Insert into wallet DB first and delete from keyfile first. Then on startup we'll have to delete stuff from wallet DB, which is less dangerous.

Both options don't sound very good :/
I'll get back to it tomorrow.

Haskell-mouse:
Maybe we shoud just check for empty wallet-db on startup and ask user about correctness of wallet-db path in case of empty wallet-db and non-empty keyfile ? 
Another option is creating backup of keyfile in case of deletion of keys on startup.

gromakovsky:
This is the thread where we discussed how to proceed with the issue: https://serokell.slack.com/archives/C9B1DAMDJ/p1536319212000100

stack-to-nix.yaml
```diff
@@ -0,0 +1 @@
+{"packages":["./ariadne/core","./ariadne/cardano","./ui/cli-app","./ui/cli-lib","./ui/qt-app","./ui/qt-lib","./ui/vty-app","./ui/vty-lib","./knit"],"resolver":"/nix/store/hqglg5hr0a95484fm0086sqyj631v6gq-term.yaml"}
```
gromakovsky: Do we need this file?

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -153,12 +155,13 @@ initializeEverything MainSettings {..}
 
     serviceAction :: IO ()
     serviceAction =
-      raceWithUpdateCheckAction $
-      cardanoAction
+      raceWithUpdateCheckAction $ cardanoAction
```
gromakovsky: `$` is redundant, btw.

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -116,7 +117,8 @@ initializeEverything MainSettings {..}
   let
     mkWalletFace :: (Doc -> IO ()) -> WalletFace
     walletInitAction :: IO ()
-    (mkWalletFace, walletInitAction) =
+    walletCheckKeys :: IO ()
+    (mkWalletFace, walletInitAction, walletCheckKeys) =
```
gromakovsky: Why not make `walletCheckKeys` part of `walletInitAction`?

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -16,6 +16,7 @@ import Text.PrettyPrint.ANSI.Leijen (Doc)
 import Ariadne.Cardano.Backend (createCardanoBackend)
 import Ariadne.Cardano.Face (CardanoEvent, CardanoFace(..), decodeTextAddress)
 import Ariadne.Config.Ariadne (AriadneConfig(..))
+-- import Ariadne.Config.Cardano (ccKeyfilePathL)
```
gromakovsky: Should be deleted.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -13,11 +13,14 @@ module Ariadne.Wallet.Backend.KeyStorage
...
        , renameSelection
        , removeSelection
+       , checkUnknownKeys
        , deriveBip44KeyPair
 
          -- * Exceptions
        , NoWalletSelection (..)
        , NoAccountSelection (..)
+       , SecretsDecodingError(..)
+-- <<<<<<< HEAD
```
gromakovsky: Should be deleted.

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -153,12 +155,13 @@ initializeEverything MainSettings {..}
...
     serviceAction =
-      raceWithUpdateCheckAction $
-      cardanoAction
+      raceWithUpdateCheckAction $ cardanoAction
 
     mainAction :: IO ()
     mainAction = do
       initAction
+      -- check if there are unknown keys in keyfile.
+      async walletCheckKeys
```
gromakovsky: `async` doesn't kill the thread it spawns when its parent thread dies. It's generally safer to use `withAsync` unless there is a specific reason to use `async`.
See https://www.notion.so/serokell/Concurrency-f0ad92790b5c4478a8f22bf5b4716fc8

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +102,31 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
 deleteRetypeMkMessage itemTypeFormat itemType = sformat
     ("Type " % itemTypeFormat % " name to confirm deletion") itemType
 
+rmUnkownHeaderMessage :: Text
+rmUnkownHeaderMessage = "Remove unknown keys"
+
+rmUnknownShureMessage :: Text
+rmUnknownShureMessage = "Make shure that you don't need keys from keyfile \
+    \(It is not backup, for example). That keys will be deleted."
```
gromakovsky: s/shure/sure

Haskell-mouse:
It is really strange (i am talking about CI), because i have "-Werror" option enabled in generated ".cabal" file an have no warnings.  Will fix now.

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: It definitely would be better, i just thought that i shoud not run async computation in walletInitAction. Now i can't say why. 
Should i move it ? It is easy.

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: ok, sorry

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: Ok, thanks. Wanted to ask about that, but forget it.

ariadne/cardano/src/Ariadne/MainTemplate.hs
gromakovsky: Yes, I also can't say why you shouldn't.

gromakovsky:
If you are using `make dev` or something similar, keep in mind that it passes `-Wwarn` to GHC.

Haskell-mouse:
updated PR

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: can't run `walletCheckKeys` in `walletInitAction` using `withAsync` - walletCheckKeys waits for confirmation from UI, but when inner action returns, thread with  `walletCheckKeys` will be killed before getting confirmation from UI. 

So - I can run it with async, or run the whole `walletInitAction` asynchronously, using `withAsync`. 
I prefer the last variant.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -452,6 +472,26 @@ renameSelection pwl WalletFace{..} walletSelRef name = do
...
+-- | This function removes unknown keys from key file on startup.
+checkUnknownKeys
+    :: PassiveWalletLayer IO
+    -> FilePath
+    -> (ConfirmationType -> IO Bool)  -- ^ What to do with unknown keys in keyfile.
+    -> IO ()
+checkUnknownKeys pwl path confirmation = do
+    keyFile <- BS.readFile path
+    roots   <- pwlGetWalletIds pwl
+    us      <- case decodeFull' @UserSecret keyFile of
```
gromakovsky: We want to hide `UserSecret` behind `Keystore`. And hide its operations behind `PassiveWalletLayer`.
I propose to add a function to `PassiveWalletLayer` which checks whether the set of keys in `Keystore` is the same as the set of wallets in acid-state DB. It shouldn't use `UserSecret` directly, it should use `toList` from `Keystore`. It can return `Set HdRootId`, for example. I. e. set of `HdRootId`s which are present in one set and absent in the other one.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -190,6 +192,14 @@ deleteHdWallet pw rootId = do
             Keystore.delete (WalletIdHdSeq rootId) (pw ^. walletKeystore)
             return $ Right ()
 
+removeKeyFromKeyStore :: PassiveWallet
+                      -> [HD.HdRootId]
+                      -> IO ()
+removeKeyFromKeyStore pw rootIds =
```
gromakovsky: Why `key` (singular) if it removes a list of keys?
Also `s` is not capital in `Keystore`, so let's stick to this naming.

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Types.hs
```diff
@@ -90,4 +90,7 @@ data PassiveWalletLayer m = PassiveWalletLayer
     , pwlLookupKeystore
           :: Kernel.HdRootId
           -> m (Maybe EncryptedSecretKey)
+    , pwlRemoveKey
```
gromakovsky: I don't like this name, how about `pwlRemoveUnknownKeys`? What I don't like:
1. `Key` is singular, though it deletes multiple keys.
2. This function shouldn't be used to delete an arbitrary key, because deletion is normally done by `pwlDeleteWallet`. It's only for a very specific case, so its name should reflect that.
Also you put it into the `internal, hopefully these will go in the future` group, but it's not really internal and we need it to go in the future.

ariadne/cardano/src/Ariadne/Wallet/Face.hs
```diff
@@ -100,6 +100,7 @@ data ConfirmationType
   = ConfirmMnemonic [Text]           -- ^ mnemonic
   | ConfirmRemove DB WalletSelection -- ^ walletDB, selection
   | ConfirmSend [ConfirmSendInfo]    -- ^ lists of outputs' info
+  | ConfirmDelUnknown Int            -- ^ number of unknown keys in keyfile
```
gromakovsky: Since this number is supposed to be small, I'd attach a list of root ids (probably as `Text`, because UI only needs `Text`) and display them instead of just a number.

ariadne/core/src/Ariadne/UIConfig.hs
gromakovsky: Also `s/backup/a backup` and `s/That/Those`.

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +102,31 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+        then "There is a key, that does not "
+        else  sformat ("There are " % int % " keys, that don't ") numberOfKeys
+      ending =
+        if numberOfKeys == 1
+        then " it ?"
+        else " them ?"
+  in beginning <> " correspond to any known wallet.\nDo you want to delete" <> ending
+
+rmUnknownRetypeMkMessage :: Text -> Text
+rmUnknownRetypeMkMessage typeMsg = "Please, write " <> typeMsg <> " if you are shure."
```
gromakovsky: `s/shure/sure` here as well

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +102,31 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+rmUnknownIntroMkMessage :: Int -> Text
+rmUnknownIntroMkMessage numberOfKeys =
+  let beginning =
+        if numberOfKeys == 1
+        then "There is a key, that does not "
+        else  sformat ("There are " % int % " keys, that don't ") numberOfKeys
+      ending =
+        if numberOfKeys == 1
+        then " it ?"
+        else " them ?"
```
gromakovsky: Also why is there a space before `?`?

ariadne/cardano/src/Ariadne/MainTemplate.hs
gromakovsky: Hm, and now I tend to think that `walletCheckKeys` is closer to service action than init action, because it should be launched after everything (including UI) is initialized, not before.
So probably your first version (returning three things from `wpMakeWallet`) was better…
But let's not call it `walletCheckKeys` here, let's call it `walletAction`, `defaultMain` doesn't need to know what exactly it does, it only needs to know that it's some action that needs to be run after initialization.

ariadne/cardano/src/Ariadne/Wallet/Face.hs
Haskell-mouse: Do you really think, that it would be useful for a user?

gromakovsky: Well, I don't like that user has to decide whether they want to delete _some keys_ when they have no idea _which keys_.
Regarding usefulness:
1. Currently it's not so useful, because the only case when user sees root ID is when they select a wallet in TUI. However, if you have two wallets with money and they are displayed in TUI and the message propose to remove some key, you can check whether corresponding root ID is one of these two wallets. If it is, you will not delete it (and it means we have a bug). If it's something different (not one of two wallets you care about), you will be sure that you can safely delete it. I think we should display IDs in GUI as well, created https://issues.serokell.io/issue/AD-486
2. I think after AD-294 and AD-168 (which I want to be done soon) wallet ID will always be passed explicitly to backend commands, so one will be able to lookup ID in history (and maybe discover that earlier they removed it).

Haskell-mouse: Oh, ok. If a user will be able to see root ID in UI, then it will be definitely useful

gromakovsky: Yes, it's already possible in TUI and will be added to GUI later.

ariadne/cardano/src/Ariadne/MainTemplate.hs
gromakovsky: @Haskell-mouse have you seen it? ↑

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Kernel.hs
```diff
@@ -114,6 +121,11 @@ passiveWalletLayerCustomDBComponent logFunction keystore acidDB = do
             , pwlGetDBSnapshot         = liftIO $ Kernel.getWalletSnapshot wallet
             , pwlLookupKeystore        = \hdrId -> liftIO $
                 Keystore.lookup (WalletIdHdSeq hdrId) (wallet ^. Kernel.walletKeystore)
+            , pwlRemoveKey             = \rootIds -> liftIO $
+                Kernel.removeKeyFromKeyStore wallet rootIds
+            , pwlGetUnknownKeys        = liftIO $ getUnknownKeys wallet
```
gromakovsky: Those functions are defined for abstract `MonadIO`, so using `liftIO` here is redundant.
Define them as `IO` or remove `liftIO` here.

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Types.hs
gromakovsky: So?

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Types.hs
```diff
@@ -90,4 +90,13 @@ data PassiveWalletLayer m = PassiveWalletLayer
...
+          :: [Kernel.HdRootId]
+          -> m ()
+
+    -- | Check keyfile.
+    , pwlGetUnknownKeys
+          :: m [Kernel.HdRootId]
+    , pwlGetWalletsWithoutSecretKeys
+          :: m ([Kernel.WalletName], [Kernel.HdRootId])
```
gromakovsky: I think it makes more sense to return a list of pairs instead of a pair of lists. It ensures that number of wallet names is the same as the number of root IDs.

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +107,41 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+rmUnknownKeysRetypeConfirm :: Text
+rmUnknownKeysRetypeConfirm = "Yes"
+
+rmBrokenWalletsHeaderMessage :: Text
+rmBrokenWalletsHeaderMessage = "Remove wallets with missing secret keys"
+
+rmBrokenWalletsIntroMkMessage :: Text
+rmBrokenWalletsIntroMkMessage = "There are wallets without corresponding secret key \
+    \in keyfile.\n Without secret keys you can do almostly nothing with such wallets. \
+    \do you want to delete them ?"
```
gromakovsky: There shouldn't be a space before `?`.

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +107,41 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+
+rmUnknownKeysRetypeConfirm :: Text
+rmUnknownKeysRetypeConfirm = "Yes"
+
+rmBrokenWalletsHeaderMessage :: Text
+rmBrokenWalletsHeaderMessage = "Remove wallets with missing secret keys"
+
+rmBrokenWalletsIntroMkMessage :: Text
+rmBrokenWalletsIntroMkMessage = "There are wallets without corresponding secret key \
+    \in keyfile.\n Without secret keys you can do almostly nothing with such wallets. \
```
gromakovsky: `almostly` is not a word.
`almost`.

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: Yes, sorry, done smth similar in another PR.

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Types.hs
Haskell-mouse: It will produce only unnecessary zipping/unzipping and it would not ensure us that number of wallet names is the same as the number of root IDs. 
Instead of that, I added checking that for every root ID in the list we have one and only one wallet.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -425,6 +433,32 @@ renameSelection pwl WalletFace{..} walletSelRef name = do
...
+-- | This function removes unknown keys from key file on startup, if user confirms that.
+checkUnknownKeys
+  :: PassiveWalletLayer IO
+  -> (ConfirmationType -> IO Bool)  -- ^ Asks user permission to delete unknown keys
+  -> IO ()
+checkUnknownKeys pwl confirmation = do
+  unknownRootIds <- pwlGetUnknownKeys pwl
+  unless (null unknownRootIds) $ do
+    let unknownRootIdsText =
+          T.dropEnd 1 . unlines $ T.drop 10 . toText . sformat build <$> unknownRootIds
```
gromakovsky: `sformat build` returns `Text`, so `toText` is most likely redundant (unless I'm missing something).

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -157,23 +158,24 @@ initializeEverything MainSettings {..}
...
-        -- It is going to call msPutBackendErrorToUI and rethrow exception, if something goes wrong
-        linkUI serviceThread $ msPutBackendErrorToUI uiFace
-        uiAction
-
+      withAsync postStartupAction $ \postStartupThread -> do
+        linkUI postStartupThread $ msPutBackendErrorToUI uiFace
+        -- Spawn backend actions in async thread, then run ui action in the main thread
+        -- This is needed because some UI libraries (Qt) insist on livng in the main thread
+        withAsync serviceAction $ \serviceThread -> do
+          -- Make backend rethrow all exceptions in main thread if something goes wrong
```
gromakovsky: It looks like you edited this comment, but it's completely unrelated to this issue, AFAIU. Maybe it was by accident.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -192,10 +192,10 @@ deleteHdWallet pw rootId = do
             Keystore.delete (WalletIdHdSeq rootId) (pw ^. walletKeystore)
             return $ Right ()
 
-removeKeyFromKeyStore :: PassiveWallet
+removeKeysFromKeyStore :: PassiveWallet
```
gromakovsky: `S` in `KeyStore` should be lowercase, because it's lowercase in all other occurrences of this word.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
```diff
@@ -174,6 +179,16 @@ instance Foldable IxSet where
 getEQ :: (Indexable a, IsIndexOf ix a) => ix -> IxSet a -> IxSet a
 getEQ ix = WrapIxSet . IxSet.getEQ ix . unwrapIxSet
 
+-- | Returns either element by it's index or
```
gromakovsky: s/it's/its

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
```diff
@@ -0,0 +1,57 @@
...
+-- And also ixset of all HdRoots
+getRootIdsFromWalletAndKeyfile
+  :: MonadIO m
+  => PassiveWallet
+  -> Keystore
+  ->  m ([HdRootId], [HdRootId], IxSet.IxSet HdRoot)
+getRootIdsFromWalletAndKeyfile pwallet keystore = do
+  snapshot <- liftIO $ getWalletSnapshot pwallet
+  let wallets = HDRead.readAllHdRoots (snapshot ^. dbHdWallets)
+  let walletRootIds = (IxSet.toList $ wallets) <&> (^. hdRootId)
```
gromakovsky:
```suggestion
  let walletRootIds = IxSet.toList wallets <&> (^. hdRootId)
```

gromakovsky: And I would use `<$>` or `map`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
```diff
@@ -0,0 +1,57 @@
...
+  :: MonadIO m => PassiveWallet -> Keystore -> m ([WalletName], [HdRootId])
+getUnknownWallets pwallet keystore = do
+  (keystoreRootIds,walletRootIds, wallets) <- getRootIdsFromWalletAndKeyfile pwallet keystore
+  let rootIDsWithMissedSecretKeys = (walletRootIds \\ keystoreRootIds)
+  let eitherWalletsWithMissedSecretKeys =
+        mapM ((flip IxSet.getTheOnly) wallets) $ rootIDsWithMissedSecretKeys
+  let walletNamesWithMissedSecretKeys = case eitherWalletsWithMissedSecretKeys of
+        Right wallet -> view hdRootName <$> wallet
+        Left  (brokenRootId, brokenWallets) ->
+                 error $ "There are more (or less) then one wallet with the same HdRootId: \n" <>
```
gromakovsky: s/then/than

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
```diff
@@ -0,0 +1,57 @@
...
+                         "Broken RootId: " <> (show brokenRootId) <> "\n" <>
```
gromakovsky: Redundant brackets.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Consistency.hs
```diff
@@ -0,0 +1,57 @@
...
+                         "Names of wallets with this id: " <>
+                         (show $ view hdRootName <$> IxSet.toList brokenWallets)
```
gromakovsky: It's better to use printing function based on `Buildable` (e. g. one from the `fmt` package).

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmRemove.hs
```diff
@@ -122,8 +188,8 @@ performContinue = whenJustM (use confirmRemoveWidgetDelRequestL) $
   where
     nameNotConfirmed delItem = case itemNameToConfirm delItem of
         Just nameValue -> do
-            confirmNameValue <- use confirmRemoveWidgetNameL
-            return $ nameValue /= confirmNameValue
+              confirmNameValue <- use confirmRemoveWidgetNameL
```
gromakovsky: Why? There were 4 spaces, you chanced it to 6 :/

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmRemove.hs
```diff
@@ -3,53 +3,101 @@ module Ariadne.UI.Vty.Widget.Dialog.ConfirmRemove
...
 
+data MessagesOnDelWidget =
+  MessagesOnDelWidget
+    { header          :: Text
+    , intro           :: Text
+    , isSureMessage   :: Text
+    , reTypeMessage   :: Text
+    , expandingMessge :: Text
+    , confirmMsg      :: Text
+    }
```
gromakovsky: Make fields strict please.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
Haskell-mouse: You are right, just forget to remove. Fixed.

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +107,41 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
 
+rmUnkownKeysHeaderMessage :: Text
+rmUnkownKeysHeaderMessage = "Remove unknown keys"
+
+rmUnknownKeysSureMessage :: Text
+rmUnknownKeysSureMessage = "Make sure that you don't need keys from keyfile \
+    \(It is not a backup, for example). Those keys will be deleted."
+
+rmUnknownKeysIntroMessage :: Text
+rmUnknownKeysIntroMessage= "There are keys, that don't \
```
artemohanjanyan: Comma should be here, I believe.

artemohanjanyan: Actually, could you check these messages in Grammarly?

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +107,41 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+    \correspond to any known wallet.\nDo you want to delete them?"
+
+rmUnknownKeysRetypeMkMessage :: Text -> Text
+rmUnknownKeysRetypeMkMessage typeMsg = "Please, write " <> typeMsg <> " if you are sure."
+
+rmUnknownKeysRetypeConfirm :: Text
+rmUnknownKeysRetypeConfirm = "Yes"
```
artemohanjanyan: In TUI it prints "...type Yes to...", I don't like the formatting. Maybe it should say "... type "Yes" to...", or "Yes" should be in bold or italic.

Haskell-mouse:
@artemohanjanyan 
Can't reproduce the bug, tried it few times and have no that new weird wallet (

ariadne/cardano/src/Ariadne/Wallet/Backend.hs
```diff
@@ -110,8 +110,11 @@ createWalletBackend walletConfig sendWalletEvent getPass voidPass = do
                     sendTx pwl this cf walletSelRef putCommandOutput getPassPhrase voidWrongPass waitUiConfirm
                 , walletBalance = getBalance pwl walletSelRef
                 }
-            initWalletAction =
-                refreshState pwl walletSelRef sendWalletEvent
+            initWalletAction = refreshState pwl walletSelRef sendWalletEvent
+
+            walletAction = do
```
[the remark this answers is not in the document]
Haskell-mouse: @gromakovsky said that our main template should not know about what exactly this function doing. At least because we can add some more functions to the walletAction in the future.

ariadne/core/src/Ariadne/UIConfig.hs
Haskell-mouse: done, will push soon.

ariadne/core/src/Ariadne/UIConfig.hs
Haskell-mouse: ok

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -192,10 +192,11 @@ deleteHdWallet pw rootId = do
             Keystore.delete (WalletIdHdSeq rootId) (pw ^. walletKeystore)
             return $ Right ()
 
-removeKeysFromKeyStore :: PassiveWallet
+removeKeysFromKeystore
+ :: PassiveWallet
```
gromakovsky: Now indentation looks very weird.

ariadne/cardano/src/Ariadne/Wallet/Backend.hs
```diff
@@ -85,7 +85,7 @@ createWalletBackend walletConfig sendWalletEvent getPass voidPass = do
             sendWalletEvent $ WalletRequireConfirm mVar confirmType
             takeMVar mVar
 
-        mkWallet cf@CardanoFace{..} = (mkWalletFace, initWalletAction)
+        mkWallet cf@CardanoFace{..} = (mkWalletFace, initWalletAction, walletAction)
```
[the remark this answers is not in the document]
gromakovsky: #364 changes this part, so I would rather avoid refactoring here.

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -7,7 +7,7 @@ module Ariadne.MainTemplate
 
 import Control.Concurrent (forkIO, myThreadId, throwTo)
 import Control.Concurrent.Async
-  (Async(..), AsyncCancelled(..), ExceptionInLinkedThread(..), race_,
+  (Async(..), AsyncCancelled(..), ExceptionInLinkedThread(..),concurrently ,race_,
```
gromakovsky: Space should be after comma, not before.

ariadne/cardano/src/Ariadne/MainTemplate.hs
gromakovsky: Everything below the line with `withAsync` should be left intact (because it's irrelevant to this PR).

ariadne/cardano/src/Ariadne/Wallet/Backend.hs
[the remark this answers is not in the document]
artemohanjanyan: Well, we can call it `postStartupAction` at least, I suppose. Again, `walletAction` is non-descriptive.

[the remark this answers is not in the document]
gromakovsky: Okay, let's call it `postStartupAction` as the middle ground. Or `postInitAction`.

ariadne/cardano/src/Ariadne/Wallet/Backend.hs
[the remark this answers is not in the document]
Haskell-mouse: It is the bug from AD-502

gromakovsky:
There's a new conflict, btw.

Haskell-mouse:
Resolved

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
```diff
@@ -125,15 +180,46 @@ itemTypeFormat :: Format r (UiDeletingItem -> r)
...
+
+unknownObjectsToggled :: Delete -> Bool -> IO ()
+unknownObjectsToggled Delete{..} expanded = case deleteUnknownObjects of
+  Just (unknownObjectsExpand,unknownObjectsScroll) -> do
+    QToolButton.setArrowType unknownObjectsExpand $ bool RightArrow DownArrow expanded
+    QWidget.setVisible unknownObjectsScroll expanded
+    QWidget.adjustSize delete
+  _ -> pass
+
+
```
gromakovsky: Extra newline.

ui/vty-app/Glue.hs
```diff
@@ -308,6 +308,10 @@ walletEventToUI = \case
       ConfirmRemoveAccount accountName ->
         UiConfirmRemove $ UiDelAccount $ Just $ unAccountName accountName
       ConfirmSend confsendInfo -> UiConfirmSend $ map toUiConfirmSendInfo confsendInfo
+      ConfirmDelUnknownKeys unknownKeys -> UiConfirmRemove $ UiDelUnknownKeys unknownKeys
+      ConfirmDelBrokenWallets walletsWithMissingKeys
+        -> UiConfirmRemove $ UiDelBrokenWallets walletsWithMissingKeys
+
```
gromakovsky: Extra newline.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
gromakovsky: Why was it marked as resolved? It's 1-space indentation here. It's not allowed.

ariadne/cardano/src/Ariadne/MainTemplate.hs
Haskell-mouse: Sorry. The change was reverted, but it seems that it returned during a rebase.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
gromakovsky: Hm, strange, I was pretty sure I was looking at the latest version last time I saw it. Ok then.

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -162,23 +163,23 @@ initializeEverything MainSettings {..}
...
       -- Spawn backend actions in async thread, then run ui action in the main thread
       -- This is needed because some UI libraries (Qt) insist on livng in the main thread
-      withAsync serviceAction $ \serviceThread -> do
-        -- Make custom link to service thread for UI apps.
-        -- It is going to call msPutBackendErrorToUI and rethrow exception, if something goes wrong
-        linkUI serviceThread $ msPutBackendErrorToUI uiFace
-        uiAction
-
+      withAsync (concurrently postStartupAction serviceAction) $ \serviceThread -> do
+          -- Make custom link to service thread for UI apps.
```
gromakovsky: Well, I still don't like that you changed the indentation here (and made it inconsistent with the rest of this function).

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -422,6 +426,31 @@ renameSelection pwl WalletFace{..} walletSelRef name = do
...
+-- | This function removes unknown keys from key file on startup, if user confirms that.
+checkUnknownKeys
+  :: PassiveWalletLayer IO
+  -> (ConfirmationType -> IO Bool)  -- ^ Asks user permission to delete unknown keys
+  -> IO ()
+checkUnknownKeys pwl confirmation = do
+  unknownRootIds <- pwlGetUnknownKeys pwl
+  unless (null unknownRootIds) $ do
+    let unknownRootIdsText =
+          T.dropEnd 1 . unlines $ T.drop 10 . sformat build <$> unknownRootIds
```
gromakovsky: This `T.drop 10` part looks pretty bad. If you want formatting different from what `build` generates, you should use a different formatter instead of removing some symbols. Now it's very fragile. E. g. someone may rename `HdRootId` and `HdRootId: ` will be changed to the new name else. Or we can decide that `Buildable` instance should be changed for another reason. Anyway, the point is that you can assume that `Buildable` instance produces something reasonable, but you can't assume that if you strip first 10 characters it will still be producing something reasonable.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -177,9 +178,14 @@ toList :: Keystore -> IO [(WalletId, EncryptedSecretKey)]
 toList (Keystore mstorage _) =
```
gromakovsky: Imo, it's better to define it in terms of `toMap`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -176,6 +177,7 @@ createWalletHdSeq pw (HasNonemptyPassphrase hasPP) name assuranceLevel esk utxoB
             if hasPP then HD.HasSpendingPassword created
                      else HD.NoSpendingPassword
 
+
```
gromakovsky: Extra newline.

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer/Kernel.hs
```diff
@@ -20,13 +20,17 @@ import qualified Ariadne.Wallet.Cardano.Kernel.Accounts as Kernel
...
+
 import qualified Ariadne.Wallet.Cardano.Kernel.Wallets as Kernel
 
 import Ariadne.Wallet.Cardano.Kernel.Bip39 (mnemonicToSeedNoPassword)
+import Ariadne.Wallet.Cardano.Kernel.Consistency
 import Ariadne.Wallet.Cardano.Kernel.DB.AcidState (DB, dbHdWallets, defDB)
 import Ariadne.Wallet.Cardano.Kernel.DB.Resolved (ResolvedBlock)
 import Ariadne.Wallet.Cardano.Kernel.Keystore (Keystore)
+
+import qualified Ariadne.Wallet.Cardano.Kernel.Keystore as Keystore (lookup)
```
gromakovsky: Why was this import moved?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Delete.hs
```diff
@@ -32,34 +37,76 @@ data Delete =
...
+      expandingMessge = "List of unknown secret keys"
+      isSureMessage = rmUnknownKeysSureMessage
+      reTypeMessage = toString $
+        rmUnknownKeysRetypeMkMessage ("<b>" <> rmUnknownKeysRetypeConfirm <> "</b>")
+      confirmMsg = rmUnknownKeysRetypeConfirm
+  in MessagesOnDelWidget{..}
+makeMessages (UiDelBrokenWallets _) =
+  let header  = toString rmBrokenWalletsHeaderMessage
+      intro = toString $ rmBrokenWalletsIntroMkMessage
+      expandingMessge = "List of wallets with a missed secret key"
```
gromakovsky: 1. Why isn't this message (and `expandingMessge = "List of unknown secret keys"` above) part of `UiConfig`?
2. s/missed/missing (and there are many places in code where `missed` is used instead of `missing`).
3. I am not entirely sure, but probably it should be `with missing secret keys`, not `with a missing secret key`.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
gromakovsky: In this particular case I would recommend doing `_fromDb . getHdRootId`.

Haskell-mouse: Hmm. What would you recommend in common? I see that my solution is not beautiful and convenient, but OTOH I don't want to change an instance, because I usually believe that it makes sense and someone made it for some reason.

gromakovsky:
```suggestion
          T.dropEnd 1 . unlines $ sformat hashHexF . _fromDb . getHdRootId <$> unknownRootIds
```

gromakovsky:
```suggestion
          T.dropEnd 1 . unlines $ sformat hashHexF . _fromDb . getHdRootId <$> unknownRootIds
```

Haskell-mouse: I understand. I mean the common case :-)

gromakovsky: In common case you can define your own formatter. If you have type `X`, don't like its `Buildable` and don't want to change it, you can define `prettierXF :: Format r (X -> r)`.

ariadne/core/src/Ariadne/UIConfig.hs
```diff
@@ -97,6 +109,47 @@ deleteRetypeMkMessage :: Format Text (delItem -> Text) -> delItem -> Text
...
+
+expandingMessageRemovingKeys :: Text
+expandingMessageRemovingKeys = "List of unknown secret keys"
+
+rmBrokenWalletsHeaderMessage :: Text
+rmBrokenWalletsHeaderMessage = "Remove wallets with missing secret keys"
+
+rmBrokenWalletsIntroMkMessage :: Text
+rmBrokenWalletsIntroMkMessage = "There are wallets without corresponding secret key \
+    \in keyfile.\n Without secret keys you can do almost nothing with such wallets. \
```
artemohanjanyan: I think that
```suggestion
    \in keyfile. Without secret keys you can do almost nothing with such wallets. \
```
is better, but :man_shrugging:

PR #301, opened 2018-09-06
Use -Weverything

Pastafarianist:
Enable `-Weverything` with exceptions as described [here](https://medium.com/mercury-bank/enable-all-the-warnings-a0517bc081c3).

Note that this should be merged after #288.

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RankNTypes, TemplateHaskell #-}
+{-# OPTIONS_GHC -fno-warn-deriving-typeable #-}
 
 -- | Acid-state database for the wallet kernel
 module Ariadne.Wallet.Cardano.Kernel.DB.AcidState (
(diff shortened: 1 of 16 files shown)


ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE RankNTypes, TemplateHaskell #-}
+{-# OPTIONS_GHC -fno-warn-deriving-typeable #-}
```
Pastafarianist: The reason for this is that `acid-state` adds a `deriving Typeable` clause in its `makeAcidic` TH wrapper.

ui/qt-lib/src/Ariadne/UI/Qt/UI.hs
```diff
@@ -23,8 +23,8 @@ class QVariantCastable a where
   fromQVariant :: (QVariant.QVariantValue v) => v -> IO a
 
 instance QVariantCastable Int where
-  toQVariant = QVariant.newWithInt . fromIntegral
-  fromQVariant = return . fromIntegral <=< QVariant.toInt
+  toQVariant = QVariant.newWithInt
+  fromQVariant = return <=< QVariant.toInt
```
goodsoft: isn't `return <=<` redundant now?

PR #304, opened 2018-09-07
[AD-285] Fix hlint warnings

Pastafarianist:
**YT issue:** https://issues.serokell.io/issue/AD-285

**Checklist:**

- [ ] Adressed HLint warnings and hints (in progress)
- [x] Tested my changes if they modify code (cosmetic changes only)

**Description:**

This uses the [ruleset](https://github.com/input-output-hk/cardano-sl/blob/develop/.hlint.yaml) we developed for Cardano.

Remaining warnings:

```
$ hlint ariadne knit ui util
knit/src/Knit/Parser.hs:56:12: Error: Parse error
Found:
      => Grammar r (Prod r Text (s, Token components) (Expr CommandId components))
    gExpr = mdo
  >     ntName <- rule $ tok _TokenName
        ntKey <- rule $ tok _TokenKey
        ntComponentsLit <- gComponentsLit @components
  

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Util.hs:49:3: Suggestion: Reduce duplication
Found:
  QBoxLayout.setStretch rowLayout 0 2
  QBoxLayout.setStretch rowLayout 1 1
  QLayout.setContentsMarginsRaw rowLayout 18 0 18 0
  QBoxLayout.addLayout layout rowLayout
  
Why not:
  Combine with ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Util.hs:61:3

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs:79:3: Warning: Reduce duplication
Found:
  widgetName <- B.getName <$> lift get
  rendered <- use logsWidgetLinesRenderedL
  total <- use logsWidgetLinesTotalL
  follow <- use logsWidgetFollowL
  whenJustM (liftBrick $ B.lookupViewport widgetName) $
    \ vp ->
      do when
           (not follow && vp ^. B.vpTop + vp ^. B.vpSize ^. _2 >= rendered)
           $
           do liftBrick $ B.invalidateCacheEntry widgetName
              when (rendered == total) $
                do liftBrick $ B.vScrollToBeginning $ B.viewportScroll widgetName
                   logsWidgetFollowL .= True
              logsWidgetLinesRenderedL .= total
  follow' <- use logsWidgetFollowL
  
Why not:
  Combine with ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs:108:5

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs:86:7: Suggestion: Reduce duplication
Found:
  liftBrick $ B.invalidateCacheEntry widgetName
  when (rendered == total) $
    do liftBrick $ B.vScrollToBeginning $ B.viewportScroll widgetName
       logsWidgetFollowL .= True
  logsWidgetLinesRenderedL .= total
  
Why not:
  Combine with ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs:115:9

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/List.hs:56:7: Suggestion: Reduce duplication
Found:
  ListWidgetState{..} <- get
  parentState <- lift $ lift get
  let items = listWidgetItemsGetter parentState
  
Why not:
  Combine with ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/List.hs:65:7

5 hints
```


goodsoft:
Duplication >>> loose abstraction

ui/qt-lib/src/Ariadne/UI/Qt.hs
```diff
@@ -93,12 +93,11 @@ qtEventSubLoop :: UiEventBQueue -> (UiEvent -> IO ()) -> Int -> IO (Either Int (
 qtEventSubLoop _ _ 0 = return $ Right ()
 qtEventSubLoop eventBQueue handler depth = do
   maybeEvent <- atomically $ tryReadTBQueue eventBQueue
-  next <- case maybeEvent of
-            Nothing -> return $ Right ()
-            Just event -> handler event >> (return $ Left $ depth - 1)
   -- This method is missing from qtah. If we notice lags here, add it to qtah and uncomment
   -- QCoreApplication.processEvents
```
goodsoft: The order matters here, I guess.
Please at least expand the comment, at least.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/List.hs
```diff
@@ -46,7 +46,7 @@ drawListWidget focused ListWidgetState{..} = do
   return $
     B.clickable widgetName $
     B.vBox $
-    map drawItem $ zip [0..] items
+    zipWith (curry drawItem) [0..] items
```
goodsoft: Much understandable wow.

ui/qt-lib/src/Ariadne/UI/Qt.hs
maksbotan: The idea was to let Qt run some of its own stuff after we run one of our handlers and before we proceed to the next one.

goodsoft: Basically, `QCoreApplication.processEvents` must be _between_ `handler event` and `return`, right?

ariadne/cardano/src/Ariadne/Config/CLI.hs
```diff
@@ -196,17 +196,16 @@ getConfig commitHash = do
 
       resolve :: FilePath -> ConfigDirectories -> FilePath -> FilePath
       resolve prefix ConfigDirectories{..} path
-        | isPrefixOf "@DATA" path = replace "@DATA" cdDataDir path
-        | isPrefixOf "@PWD" path = replace "@PWD" cdPWD path
+        | "@DATA" `isPrefixOf` path = replace "@DATA" cdDataDir path
```
int-index: The check is whether `@DATA` is a prefix, but `replace` modifies *every* occurrence. Is this intended?

ui/qt-lib/src/Ariadne/UI/Qt.hs
maksbotan: Yes. But now I'm confused about when haskell will actually call these two functions (handler and processEvents).

ariadne/cardano/src/Ariadne/Config/CLI.hs
gromakovsky: Yes, we want to replace every occurrence of `@DATA`. We check whether it's a prefix because if it's a prefix we treat this part as absolute, otherwise we treat it as relative.

gromakovsky:
> This uses the ruleset we developed for Cardano.

Not we, some other people. You can see [here](https://github.com/input-output-hk/cardano-sl/pull/2745/files) how it was changed.
I think you should put the ruleset into the repository, so that others can use it as well.

gromakovsky:
The biggest difference is in `Universum` specific rules (they are not present). Up-to-date rules can be found in https://github.com/serokell/universum/blob/master/.hlint.yaml.
But I think there are too many rules and we should pick only the most useful ones selectively.

gromakovsky:
Also there is a challenge to apply them only to packages which use `Universum`, but not `knit`.

gromakovsky:
Probably we need a script to do all of this (like https://github.com/input-output-hk/cardano-sl/blob/ffe2a9ea0073df0b0259d1ce91ea8ca8e716662d/scripts/haskell/lint.sh).

Pastafarianist:
Applying different rules to `knit` and everything else should be easy, since invoking `hlint` is just passing a list of packages to it.

gromakovsky:
Still, I'd like to have a script to do it.

ariadne/cardano/src/Ariadne/Config/CLI.hs
```diff
@@ -196,17 +196,16 @@ getConfig commitHash = do
...
+        | "@PWD" `isPrefixOf` path = replace "@PWD" cdPWD path
```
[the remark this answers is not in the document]
Pastafarianist: It does. I'd argue that the infix form of `isPrefixOf` is more readable.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/List.hs
Pastafarianist: I agree that it is less transparent, but my own impression of simplicity does not always match that of others, so I wanted to see some opinions. I will revert this.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+- arguments: [-XTypeApplications, -XRecursiveDo, --cpp-define=CONFIG=dev]
```
gromakovsky: > --cpp-define=CONFIG=dev

wut

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
```
gromakovsky: This paragraph is likely useless.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
...
+# - ignore: {name: Use let}
+# - ignore: {name: Use const, within: SpecialModule} # Only within certain modules
+
+
+# Define some custom infix operators
+# - fixity: infixr 3 ~^#^~
+
+
+# To generate a suitable file for HLint do:
+# $ hlint --default > .hlint.yaml
```
gromakovsky: This paragraph is likely useless as well.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
...
+- ignore: {name: Unused LANGUAGE pragma}
+- ignore: {name: Use newtype instead of data}
+# Rules not found in old HLint.hs file (prior to HLint 2.0)
+# Added when we made the change.
+- ignore: {name: Redundant lambda}
+- ignore: {name: Use section}
+
+# Specify additional command line arguments
+#
+# - arguments: [--color, --cpp-simple, -XQuasiQuotes]
```
gromakovsky: It seems useless, because we are using this feature anyway.
In fact, I'd remove all these suggestions and leave a link to e. g. https://github.com/ndmitchell/hlint#customizing-the-hints, but that's up to you.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+- arguments: [-XTypeApplications, -XRecursiveDo, --cpp-define=CONFIG=dev]
+
+- ignore: {name: Redundant do}
```
gromakovsky: Imo, it'd be better to use our version as base, not IOHK's version, i. e. this one: https://github.com/input-output-hk/cardano-sl/pull/2745/files#diff-c1b5b6b6e840f412ea68b4acec74d3ec
The reason is: `ignore`s are grouped there and have comments which make review easier.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs
```diff
@@ -61,12 +60,12 @@ drawLogsWidget LogsWidgetState{..} = do
...
         drawLogMessage (LogMessage message) =
           V.vertCat $
-          csiToVty attr <$> Text.lines message
+          csiToVty attr <$> lines message
         img =
           (if logsWidgetFollow then V.cropTop height else identity) $
           V.vertCat $
           reverse $
-          fmap drawLogMessage $
+          drawLogMessage <$>
```
Pastafarianist: This one is debatable. I'm thinking it should not be changed, but I'm willing to listen to others' opinion.

.hlint-universum.yaml
```diff
@@ -0,0 +1,414 @@
...
+    note: "'sortWith' will be faster here because it doesn't do caching" }
+- warn: { lhs: sortOn (Down . fst), rhs: Universum.sortWith (Down . fst),
+    note: "'sortWith' will be faster here because it doesn't do caching" }
+- warn: { lhs: sortOn (Down . snd), rhs: Universum.sortWith (Down . snd),
+    note: "'sortWith' will be faster here because it doesn't do caching" }
+
+- warn: { lhs: fmap concat (mapM f s), rhs: Universum.concatMapM f s }
+- warn: { lhs: concat <$> mapM f s, rhs: Universum.concatMapM f s }
+
+# Removed for now since we don't want to make people use (some of) our ad-hoc stuff.
```
gromakovsky: I'd remove it completely (i. e. this whole commented out stuff), that's indeed ad-hoc and not really useful.

.hlint-universum.yaml
```diff
@@ -0,0 +1,414 @@
...
+- warn: { name: "liftIO is not needed", lhs: liftIO (atomicModifyIORef x y), rhs: Universum.atomicModifyIORef x y,
+    note: "If you import 'atomicModifyIORef' from Universum, it's already lifted" }
+
+- warn: { name: "liftIO is not needed", lhs: liftIO (atomicModifyIORef' x y), rhs: Universum.atomicModifyIORef' x y,
+    note: "If you import 'atomicModifyIORef'' from Universum, it's already lifted" }
+
+- warn: { name: "liftIO is not needed", lhs: liftIO (atomicWriteIORef x y), rhs: Universum.atomicWriteIORef x y,
+    note: "If you import 'atomicWriteIORef' from Universum, it's already lifted" }
+
+# others
```
gromakovsky: Did you take some old version of rules? Some of these rules are not present in latest `master`. I'd rather use latest rules, because we'll start using newest `universum` at some point where these rules don't make sense.

.hlint-universum.yaml
```diff
@@ -0,0 +1,414 @@
...
+- warn: { lhs: unlessM (not <$> x), rhs: whenM x }
+
+# Oh boy, we sure have many ways of spelling “pure ()”. Also I checked and
+# HLint isn't smart enough to see reordered case branches.
+- warn: { lhs: (case m of Just x -> f x; Nothing -> pure ()  ), rhs: Universum.whenJust m f }
+- warn: { lhs: (case m of Just x -> f x; Nothing -> return ()), rhs: Universum.whenJust m f }
+- warn: { lhs: (case m of Just x -> f x; Nothing -> pass     ), rhs: Universum.whenJust m f }
+- warn: { lhs: (case m of Nothing -> pure ()  ; Just x -> f x), rhs: Universum.whenJust m f }
+- warn: { lhs: (case m of Nothing -> return (); Just x -> f x), rhs: Universum.whenJust m f }
+- warn: { lhs: (case m of Nothing -> pass     ; Just x -> f x), rhs: Universum.whenJust m f }
```
gromakovsky: I think that sometimes `case` version is better, mostly when `m` and `f` are big.

.hlint-universum.yaml
```diff
@@ -0,0 +1,414 @@
...
+# Unfortunately, these are often bad because they remove a variable name
+# (which usually clarifies things):
+# - suggest: { lhs: (do x <- m; when x a), rhs: Universum.whenM m a }
+# - suggest: { lhs: (do x <- m; unless x a), rhs: Universum.unlessM m a }
+
+- warn: { lhs: whenM (not <$> x), rhs: unlessM x }
+- warn: { lhs: unlessM (not <$> x), rhs: whenM x }
+
+# Oh boy, we sure have many ways of spelling “pure ()”. Also I checked and
+# HLint isn't smart enough to see reordered case branches.
```
gromakovsky: Rules in `master` have a comment with this link https://github.com/ndmitchell/hlint/issues/434

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Logs.hs
gromakovsky: I like `<$>` version more, but I think `fmap` should be acceptable.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
...
+- ignore: { name: Use fmap }
+- ignore: { name: Use foldl }
+- ignore: { name: "Use :" }
+- ignore: { name: Use ++ }
+- ignore: { name: Use || }
+- ignore: { name: Use && }
+- ignore: { name: 'Use ?~' }
+- ignore: { name: Use <$> }
+
+# There's no 'head' in Universum
```
gromakovsky: `universum`-specific.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
...
+- ignore: { name: Use head }
+
+# Sometimes [Char] is okay (if it means "a set of characters")
+- ignore: { name: Use String }
+
+# We have 'whenJust' for this
```
gromakovsky: `universum`-specific.

.hlint.yaml
```diff
@@ -0,0 +1,93 @@
...
+# Sometimes TemplateHaskell is needed to please stylish-haskell
+- ignore: { name: Unused LANGUAGE pragma }
+
+# Some 'data' records will be extended with more fields later,
+# so they shouldn't be replaced with 'newtype' blindly
+- ignore: { name: Use newtype instead of data }
+
+
+###########################################################################
+# Hints with 'id' should use 'identity'
```
gromakovsky: `universum`-specific and doesn't even make any sense nowadays, because `identity` is deprecated.

.hlint-universum.yaml
Pastafarianist: I took the version you linked in https://github.com/serokell/ariadne/pull/304#discussion_r216774817.

.hlint-universum.yaml
Pastafarianist: I don't think that the size of `f` matters, since it is used like so:
```
whenJust m $ do
    <make your f as big as you like>
```
as for `m`, I don't see much difference between `case m of` and `whenJust m $ do`. The latter is slightly longer, but that's just a few characters.

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+- warn: {lhs: Data.Text.pack (show x), rhs: Universum.show x}
+- warn: {lhs: Data.Text.Lazy.pack (show x), rhs: Universum.show x}
+
+- warn: {lhs: Control.Exception.evaluate, rhs: evaluateWHNF}
+- warn: {lhs: Control.Exception.evaluate (force x), rhs: evaluateNF x}
+- warn: {lhs: Control.Exception.evaluate (x `deepseq` ()), rhs: evaluateNF_ x}
+- warn: {lhs: void (evaluateWHNF x), rhs: evaluateWHNF_ x}
+- warn: {lhs: void (evaluateNF x), rhs: evaluateNF_ x}
+
+## Monad.Trans
```
gromakovsky: Okay, let's discuss. I think this stuff is too non-standard and we shouldn't force people to use it.

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+- warn: {lhs: flip runReaderT , rhs: usingReaderT }
+- warn: {lhs: flip runReader  , rhs: usingReader }
+- warn: {lhs: flip runStateT  , rhs: usingStateT }
+- warn: {lhs: flip runState   , rhs: usingState }
+- hint: {lhs: fst <$> usingStateT s st, rhs: evaluatingStateT s st}
+- hint: {lhs: fst (usingState s st), rhs: evaluatingState s st}
+- hint: {lhs: snd <$> usingStateT s st, rhs: executingStateT s st}
+- hint: {lhs: snd (usingState s st), rhs: executingState s st}
+
+## Containers
```
gromakovsky: `keys`, `elems`, `toPairs` are non-standard as well, don't wanna force to use them (but less than the previous block).

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+- warn: { lhs: sortOn snd, rhs: Universum.sortWith snd
+        , note: "'sortWith' will be faster here because it doesn't do caching" }
+- warn: { lhs: sortOn (Down . fst), rhs: Universum.sortWith (Down . fst)
+        , note: "'sortWith' will be faster here because it doesn't do caching" }
+- warn: { lhs: sortOn (Down . snd), rhs: Universum.sortWith (Down . snd)
+        , note: "'sortWith' will be faster here because it doesn't do caching" }
+
+- warn: {lhs: map fst &&& map snd, rhs: unzip}
+
+- hint: {lhs: "foldl' (flip f)", rhs: "flipfoldl' f"}
```
gromakovsky: :unamused:

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+
+- warn: {lhs: f >>= guard, rhs: guardM}
+- warn: {lhs: guard =<< f, rhs: guardM}
+
+- hint: {lhs: fmap (fmap f) x, rhs: f <<$>> x}
```
gromakovsky: I don't like `<<$>>` :man_shrugging:

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+- warn: {lhs: maybe (Right r) Left, rhs: maybeToLeft}
+
+- warn: {lhs: fromMaybe mempty, rhs: maybeToMonoid}
+- warn: {lhs: "m ?: mempty", rhs: maybeToMonoid m}
+
+
+- hint: {lhs: pure (), rhs: pass}
+- hint: {lhs: return (), rhs: pass}
+
+# Probably will be reduced when function equality is done:
```
gromakovsky: Regarding this `when*` stuff, let's give it a try, but maybe I'll want to remove these rules later.

.hlint-universum.yaml
```diff
@@ -0,0 +1,1050 @@
...
+- warn: { name: "Use 'getLine' from Universum"
+        , lhs: Data.Text.IO.getLine, rhs: Universum.getLine }
+- warn: { name: "Use 'readFile' from Universum"
+        , lhs: Data.Text.IO.readFile, rhs: Universum.readFile }
+- warn: { name: "Use 'writeFile' from Universum"
+        , lhs: Data.Text.IO.writeFile, rhs: Universum.writeFile }
+- warn: { name: "Use 'appendFile' from Universum"
+        , lhs: Data.Text.IO.appendFile, rhs: Universum.appendFile }
+
+## Unsafe
```
gromakovsky: I wouldn't enforce that.

.hlint-universum.yaml
Pastafarianist: I probably agree with you on this one. _However_, this will have to be reconciled with the fact that these functions still remain in Universum. Perhaps we should deprecate them?

.hlint-universum.yaml
Pastafarianist: I disagree on this one, because unlike `{using,evaluating,executing}{Reader,State}{,T}` these functions provide a consistent interface to multiple containers instead of increasing cognitive burden for remembering names.

.hlint-universum.yaml
Pastafarianist: Agree. Deprecation?

.hlint-universum.yaml
Pastafarianist: I do :man_shrugging:

.hlint-universum.yaml
Pastafarianist: I think these are OK because it is very easy to infer from their names what they do.

.hlint-universum.yaml
Pastafarianist: People seem to [like](https://serokell.slack.com/archives/C9B1DAMDJ/p1537546071000100) the idea.

.hlint-universum.yaml
gromakovsky: I generally don't think that we should force to much. If some function exists, we shouldn't necessarily require using it whenever possible or deprecate it instead. Too severe limitations are usually bad.
If some person likes `usingReaderT`, I'd let them use it, but if they don't, `flip runReaderT` should be acceptable as well. I don't think that one of these options is so much better than another one.

PR #307, opened 2018-09-08
[AD-335] Introduce hpack

goodsoft:
**YT issue:** https://issues.serokell.io/issue/AD-335

**Checklist:**

  - [x] [README](README.md)

**Description:**
Much less boilerplate!


hpack/definitions.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - TupleSections
+        - TypeApplications
+        - TypeFamilies
+        - TypeOperators
+        - UndecidableInstances
+        - ViewPatterns
+
+    - &dependencies
+        - base >= 4.7 && < 5
+        - universum
```
gromakovsky: AFAIU, because of that `knit` depends on `universum` and I suspect @int-index may not like it (beside not liking `hpack` in general :sweat_smile:).

goodsoft: That one is fixed in the next PR, will probably port it to this one

goodsoft: Fixed with an override in the local `package.yaml`

gromakovsky:
Is Cabal 3.0 going to be released in 2018? If not, why should we care about transition to it _now_? Not using `hpack` complicates some other things (e. g. maintaining proper list of warnings, default extensions) _now_.

int-index:
> please elaborate on these complications.

The removal of .cabal files is problematic, Cabal 3.0 will definitely keep using them. In particular:

* factoring out parts of the configuration in ways not supported by 
https://cabal.readthedocs.io/en/latest/developing-packages.html#common-stanzas
* using auto-discovery of .hs-files

> Is Cabal 3.0 going to be released in 2018?

I hope so, but if you’re impatient the version from `master` works reasonably well, I’m using it.

goodsoft:
> The removal of .cabal files is problematic, Cabal 3.0 will definitely keep using them

I once again remind you that .cabal files are not removed, but rather become a generated asset.
So, if at some point hpack becomes unable to generate a .cabal file we need, we just remove `package.yaml` and start modifying the previously generated .cabal file ourselves.

So, switching to Cabal 3.0 can become _slightly_ problematic only in the following cases:
1. They make breaking syntax changes _and_ hpack doesn't follow. See the paragraph above for an _easy_ fix.
2. They introduce their own way of DRY. Well, in this case we will still need to switch to it, but it will already be easier, because the common things are already factored out by this PR.

In any case, I still don't see any substantial problems in introducing hpack, but rather a huge benefit.

If you still insist on hpack being a problem, please provide a link to a summary of breaking changes in Cabal 3.0 and/or description of how it will solve the problem of configuration duplication and module auto-discovery.

gromakovsky:
AFAIK, `hpack` is used at least in `disciplina` (and probably in some other our projects) and everything works. Why are there difficulties with this particular project, is it somewhat special?

PR #310, opened 2018-09-11
[AD-64] Password Manager

pasqu4le:
**YT issue:** https://issues.serokell.io/issue/AD-64

**Checklist:**

- [] Updated docs if necessary
  - [x] [README](README.md) (not necessary)
  - [x] [Configuration documentation](docs/configuration.md) (not necessary)

**Description:**
Password Manager implementation as well as UI and backend changes to use it.


ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+    , confirmButton :: QPushButton.QPushButton
+    }
+
+data InsertPasswordResult
+    = InsertPasswordCanceled
+    | InsertPasswordAccepted (Maybe Text)
+
+makeLensesWith postfixLFields ''InsertPassword
+
+createPasswordField :: Text -> IO (QHBoxLayout.QHBoxLayout, QLineEdit.QLineEdit)
```
maksbotan: This is copy-pasted from `Dialogs/NewWallet.hs`. Please factor out this functiont to `Dialogs/Util.hs`

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+createPasswordField placeholder = do
+    layout <- QHBoxLayout.new
+    field <- QLineEdit.new
+    QLineEdit.setEchoMode field QLineEdit.Password
+    QLineEdit.setPlaceholderText field $ toString placeholder
+
+    visibleButton <- QPushButton.new
+    void $ setProperty visibleButton ("styleRole" :: Text) ("passwordVisibilityToggle" :: Text)
```
maksbotan: This style (`passwordVisibilityToggle`) is applied only to buttons in `#newWalletDialog`. I believe you should change QSS to `QDialog QPushButton[styleRole="passwordVisibilityToggle"]` (see `ui/qt-lib/resources/stylesheet.qss`)

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+    return (layout, field)
+
+initInsertPassword :: IO InsertPassword
+initInsertPassword = do
+    insertPassword <- QDialog.new
+    QObject.setObjectName insertPassword ("insertPasswordDialog" :: String)
+    QWidget.setWindowTitle insertPassword ("Insert Password" :: String)
+    QWidget.adjustSize insertPassword
+
+    layout <- createLayout insertPassword
```
maksbotan: Looks like dialog lacks a header label, see `createHeader` in `Dialogs/Util.hs`.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+
+    hasPasswordLabel <- QLabel.newWithText $ toString passwordLabelText
+    QLabel.setWordWrap hasPasswordLabel True
+    QWidget.setMinimumSizeRaw hasPasswordLabel 600 30
+    hasPassword <- QCheckBox.new
```
maksbotan: Can you explain why is this needed here? I think it's kinda silly when password manager requests a password when a wallet does not have one.

Also, I was unable to create new address in paswordless wallet - a dialog runs, I de-check this checkbox, click confirm and then get `Address generation failed due to incorrect passphrase`.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+initInsertPassword :: IO InsertPassword
+initInsertPassword = do
+    insertPassword <- QDialog.new
+    QObject.setObjectName insertPassword ("insertPasswordDialog" :: String)
+    QWidget.setWindowTitle insertPassword ("Insert Password" :: String)
+    QWidget.adjustSize insertPassword
+
+    layout <- createLayout insertPassword
+
+    hasPasswordLabel <- QLabel.newWithText $ toString passwordLabelText
```
maksbotan: Oh, and by the way, use `createCheckBox` from `Util.hs` for this.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,153 @@
...
+    QWidget.setSizePolicyRaw hasPassword Maximum Preferred
+
+    addRow layout hasPasswordLabel hasPassword
+
+    (passwordWidget, passwordLayout) <- createSubWidget
+    passwordLabel <- QLabel.newWithText ("<b>PASSWORD</b>" :: String)
+    (passwordFieldLayout, password) <- createPasswordField "password"
+
+    addRowLayout passwordLayout passwordLabel passwordFieldLayout
+    addSeparator passwordLayout
```
maksbotan: I believe the separator is not needed here.

pasqu4le:
I fixed all (I think) HLint that were in the files I modified (even when I did not introduce them). I forgot about #304 

> Can you explain why is this needed here? I think it's kinda silly when password manager requests a password when a wallet does not have one.
>
> Also, I was unable to create new address in paswordless wallet - a dialog runs, I de-check this checkbox, click confirm and then get Address generation failed due to incorrect passphrase.

On slack there is still an open discussion on whether to ask a password for a password-less wallet or not. There is also another discussion on if there should be a difference between no-password (represented as a `Nothing`) and an empty password (a `""`).

I made the latest commit to try both case: as it is now you can create a new paswordless wallet and it will work as well (that uses `""`), but breaks the previous ones (the `Nothing`s). If you avoid this commit you should have no problem using the paswordless wallet you had before.

I will solve the other things!

maksbotan:
> If you avoid this commit you should have no problem using the paswordless wallet you had before.

Well, this looks like user-visible breakage. Don't know if it's severe though. @gromakovsky ?

maksbotan:
Can you rebase it on current master in the meantime?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
maksbotan: I just realized that that function will place the checkbox _left_ of the label. If you still need the checkbox, please try to refactor the function to support both cases.

pasqu4le: @maksbotan `NewWallet` also does not use `createCheckBox`, probably because it also has a checkbox on the right. Should I fix that too?

maksbotan: AFAIR, that's the only place right checkbox was needed at the moment, so I did not bother to generalize it. Would be great if you fixed it.

However, I'd wait for the final decision whether that checkbox is needed.

pasqu4le: Already changed `createCheckBox` to take another parameter, will change `NewWallet` and push it. If there are problems I will revert it back later.

ariadne/cardano/src/Ariadne/Wallet/Knit.hs
```diff
@@ -147,11 +149,12 @@ instance (Elem components Wallet, Elem components Core, Elem components Cardano)
         { cpName = newWalletCommandName
         , cpArgumentPrepare = identity
         , cpArgumentConsumer = do
-            passPhrase <- getPassPhraseArg
+            walletRef <- getWalletRefArgOpt
```
goodsoft: Oh wait, we don't have any wallet ref yet, when we are _creating_ a wallet

ariadne/cardano/src/Ariadne/Wallet/Knit.hs
```diff
@@ -161,14 +164,15 @@ instance (Elem components Wallet, Elem components Core, Elem components Cardano)
         { cpName = restoreCommandName
         , cpArgumentPrepare = identity
         , cpArgumentConsumer = do
-            passPhrase <- getPassPhraseArg
+            walletRef <- getWalletRefArgOpt
```
goodsoft: Same here

ui/cli-lib/src/Ariadne/UI/Cli.hs
```diff
@@ -11,61 +11,89 @@ import Data.Text (strip, unlines)
...
-createAriadneUI :: ComponentM (UiFace, UiAction)
-createAriadneUI = buildComponent_ "UI-CLI" $ do
-    printActionVar :: MVar PrintAction <- newEmptyMVar
-    let uiFace = mkUiFace printActionVar
-    return (uiFace, runUI printActionVar uiFace)
+data AppState = AppState
+    { loopDecision :: !LoopDecision
+    -- ^ tracks if it has to stop the loop or keep going
+    , passwordMode :: !(Maybe (WalletReference, CE.Event))
+    -- ^ tracks if it's in passeword mode and the info it need to communicate
```
goodsoft: Typo

ui/cli-lib/src/Ariadne/UI/Cli.hs
```diff
@@ -76,13 +104,11 @@ isHelpCommand t = strip t `elem` ["help"]
...
 outputTextLn = Haskeline.outputStrLn . toString
 
-getInputTextLine :: Text -> Repl (Maybe Text)
-getInputTextLine = (fmap.fmap) toText . Haskeline.getInputLine . toString
+getInputTextLine :: Repl (Maybe Text)
+getInputTextLine = getUsingText Haskeline.getInputLine "knit> "
 
-data LoopDecision = Continue | Stop
+getTextPassword :: Repl (Maybe Text)
+getTextPassword = getUsingText (Haskeline.getPassword (Just '*')) "password> "
```
goodsoft: Please make the prompt `Password: `, like it's usually in shells.
Also, I think we should disable asterisks in the same vein.

ui/qt-lib/ariadne-qt-lib.cabal
```diff
@@ -25,7 +25,8 @@ library
   hs-source-dirs:
       src
   build-depends:
-      ariadne-core
+      ariadne-cardano
```
goodsoft: This is definitely wrong, because UI lib _must not_ depend on concrete currency backend.

ui/cli-app/Glue.hs
```diff
@@ -46,8 +51,9 @@ knitFaceToUI
      )
   => UiFace
   -> KnitFace components
+  -> PutPassword WalletReference
   -> UiLangFace
-knitFaceToUI UiFace{..} KnitFace{..} =
+knitFaceToUI UiFace{..} KnitFace{..} _ =
```
goodsoft: Why add it here, if we don't use it anyway?

ui/cli-lib/ariadne-cli-lib.cabal
```diff
@@ -18,8 +18,11 @@ library
   hs-source-dirs:
       src
   build-depends:
-      ansi-wl-pprint
+      ariadne-cardano
```
goodsoft: No Cardano here too, please.

ui/cli-lib/src/Ariadne/UI/Cli/Face.hs
```diff
@@ -41,6 +45,11 @@ data UiLangFace = forall err expr. UiLangFace
 -- capture /what the UI can handle/, not what the backends can generate.
 data UiEvent
   = UiCommandEvent UiCommandId UiCommandEvent
+  | UiPasswordEvent UiPasswordEvent
+
+-- | Ui event triggered by the password manager
+data UiPasswordEvent
+  = UiPasswordRequest WalletReference CE.Event
```
goodsoft: Type variable, I suppose. UI definitely shouldn't care, what kind of references does the backend use.

ui/vty-lib/ariadne-vty-lib.cabal
```diff
@@ -18,12 +18,14 @@ library
   hs-source-dirs:
       src
   build-depends:
-      ansi-wl-pprint
+      ariadne-cardano
```
goodsoft: And here (also ordering)

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
```diff
@@ -214,8 +222,8 @@ data UiRestoreWalletArgs = UiRestoreWalletArgs
   , urwaFull :: !Bool
   }
 
-data UiRenameArgs = UiRenameArgs
-  { uraName :: !Text
+newtype UiRenameArgs = UiRenameArgs
+  { uraName :: Text
```
goodsoft: :unamused:

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/Edit.hs
```diff
@@ -37,6 +37,7 @@ data EditWidgetState p =
     , editWidgetAttr :: !B.AttrName
     , editWidgetCharTransform :: !(Char -> Char)
     , editWidgetCharAttr :: !(Maybe (p -> (Int, Int) -> B.AttrName))
+    , editWidgetCharTransformP :: !(Maybe (p -> Char -> Char))
```
goodsoft: Why do we need this?

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Repl.hs
```diff
@@ -162,7 +181,7 @@ drawReplWidget focus ReplWidgetState{..} = do
               Just mkImg -> mkImg ! #def_attr defAttr ! #width width
           ]
 
-    inputPrompt = "knit> "
+    inputPrompt = if isJust replWidgetPasswordMode then "password> " else "knit> "
```
goodsoft: `Password: ` here too pls.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Repl.hs
```diff
@@ -66,15 +77,23 @@ initReplWidget uiFace langFace historyFace fullsizeGetter =
...
 
-    addWidgetChild WidgetNameReplInput $
-      initBaseEditWidget (widgetParentLens replWidgetCommandL) "default" id (Just $ widgetParentGetter spanAttrs) EnterWithBackslash
+    addWidgetChild WidgetNameReplInput $ initBaseEditWidget
+        (widgetParentLens replWidgetCommandL)
+        "default"
+        id
+        (Just $ widgetParentGetter spanAttrs)
+        (Just $ widgetParentGetter charModifier)
+        EnterWithBackslash
```
goodsoft: I would add a separate password widget, which would be shown only in the password mode.
Also, it should be automatically focused, when the password is requested.
I suppose, it should be done in the same manner as `MenuWidget`, i.e. defined in `App` and having a separate mode, so that all events are dispatched to it.

pasqu4le:
Ok, first thing I am going to do is add `CheckboxPosition` and capitalize the header as requested by @maksbotan.

The removal of unrelated Hlint fixes will follow.
Also, `newType`s were suggested by Hlint, but they cannot have strictness annotation, I assumed that was better (wrongly?)

I put the `ariadne-cardano` dependency there only for `WalletReference`, I'll remove it and use a type variable.

`WalletRefSelection` will be used directly instead of using `getWalletRefArgOpt` when creating/restoring a wallet. Anyway this might be a problem, because `WalletRefSelection` could already be in the `Map`.

`editWidgetCharTransformP` was so that the `Edit` widget could know what character transformation function to use from it's parent widget. Since I will add a new widget this will be removed. Also, this new widget will take the space of the Edit widget in the Repl when it is in password mode, did I get this correctly?

goodsoft:
> Also, this new widget will take the space of the Edit widget in the Repl when it is in password mode, did I get this correctly?

No, It should be under it, I think.
Or maybe go full derp and display it as a popup in front of the whole UI, as a separate Brick layer, as it will be modal anyway.

pasqu4le:
I suppose modal would not be that bad, even when getting the focus it might be not very visible down there.

ui/cli-lib/src/Ariadne/UI/Cli/Face.hs
pasqu4le: If I use `data UiPasswordEvent walletRef = UiPasswordRequest walletRef CE.Event` then `UiEvent` will also need it. Are you sure with the type variable?

ariadne/core/src/Ariadne/UX/PasswordManager.hs
```diff
@@ -0,0 +1,104 @@
...
+
+data WalletId
+    = WalletIdTemporary
+    -- ^ Temporary ID: a password saved with this ID will be deleted the
+    -- first time it's read
+    | WalletIdSelected
+    -- ^ identifies the wallet currently selected
+    | WalletIdByUiIndex !Word
+    -- ^ id based on the Ui index
+    -- TODO: maybe by hash?
```
gromakovsky: That's AD-294, please add a reference to this comment.

docs/usage-tui.md
```diff
@@ -75,7 +75,7 @@ Ariadne using the 12 word mnemonic.
 When you select a wallet or an account, you can see its balance and send 
 transactions using the form provided. You must fill at least one receiver's 
 address. To add or remove address lines, use `[ - ]` / `[ + ]` buttons on
-the right. If the wallet has a passphrase, you need to specify that as well. 
+the right. You will be asked for a passphrase, even if you chose an empty one.
```
goodsoft: Wait a second, will it ask for a password, when I use the send form (which already has a password input)?

pasqu4le: Nope, it will only ask if you put the command in the `Repl` (in TUI).

pasqu4le: Oh no, I just checked, it does ask you for one anyway, will fix this.

pasqu4le: I was asking myself why I removed the password from `UiSendArgs` if there was a field in the form and then I remembered: there is no field anymore.

You can see it was removed some time ago, in the second commit, I suppose it was discussed, but I do not remember. Should I put it back in?

goodsoft: Well, I think that existing forms with password fields should work as before.
That is, they should preemptively provide a password to the password manager.
The password should be explicitly asked only when you type a command in REPL.

ui/qt-app/Glue.hs
```diff
@@ -321,3 +332,11 @@ historyToUI ch = UiHistoryFace
...
   , historyPrevCommand = toPrevCommand ch
   }
+
+----------------------------------------------------------------------------
+-- Glue between the Password Manager and Vty frontend
+----------------------------------------------------------------------------
+
+putPasswordEventToUI :: UiFace -> WalletId -> CE.Event -> IO ()
+putPasswordEventToUI UiFace{..} walletId cEvent = putUiEvent . UiPasswordEvent $
+    UiPasswordRequest walletId cEvent
```
maksbotan: New line again.

ui/qt-app/Glue.hs
```diff
@@ -109,15 +122,13 @@ knitFaceToUI UiFace{..} KnitFace{..} =
                 ]
             ]
           )
-      UiRestoreWallet name password mnemonic full -> do
+      UiRestoreWallet name _ mnemonic full -> do
```
maksbotan: Maybe 'password' argument can be dropped from `UiRestoreWallet` completely?

ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
```diff
@@ -146,7 +154,7 @@ data UiHistoryFace =
     }
 
 -- API for the UI.
-data UiFace = UiFace
+newtype UiFace = UiFace
```
maksbotan: Stale hlint fix?

ui/qt-lib/src/Ariadne/UI/Qt/MainWindow.hs
```diff
@@ -82,15 +87,23 @@ handleMainWindowEvent langFace = \case
     magnify replL $ handleReplEvent commandId result
   UiCommandResult commandId commandResult -> case commandResult of
     UiSendCommandResult result ->
-      magnify walletL $ handleWalletEvent langFace $ WalletSendCommandResult commandId result
+      magnify walletL $ handleWalletEvent langFace putPass $
```
maksbotan: I don't like this formatting noise, but maybe let's leave it.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/InsertPassword.hs
```diff
@@ -0,0 +1,118 @@
...
+
+    header <- QLabel.newWithText . toString $ T.toUpper headerText
+    addHeader layout header
+
+    hasPassword <- createCheckBox layout CheckboxOnRight passwordLabelText
+    QAbstractButton.setChecked hasPassword True
+
+    (passwordWidget, passwordLayout) <- createSubWidget
+    passwordLabel <- QLabel.newWithText ("<b>PASSWORD</b>" :: String)
+    (passwordFieldLayout, password) <- createPasswordField "password"
```
maksbotan: Here's a slight problem with this. AFAIU, when creating new wallet, password manager would ask to provide a password and set it as wallet's password. However, in this case we'd like user to enter password twice, as confirmation.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/NewWallet.hs
```diff
@@ -135,26 +103,16 @@ initNewWallet = do
...
   QBoxLayout.addWidget layout walletMnemonicWidget
   QWidget.hide walletMnemonicWidget
 
   addSeparator layout
 
-  hasPasswordLabel <- QLabel.newWithText $ toString passwordLabelText
-  QLabel.setWordWrap hasPasswordLabel True
-  QWidget.setMinimumSizeRaw hasPasswordLabel 600 30
-  hasPassword <- QCheckBox.new
+  hasPassword <- createCheckBox layout CheckboxOnRight passwordLabelText
```
maksbotan: Everything related to passwords in this dialog is now useless and should probably be removed. This dialog creates new wallets with `UiRestoreWallet` command and you now ignore its password field.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Util.hs
```diff
@@ -87,17 +93,26 @@ createSubWidget = do
...
-createCheckBox layout labelText = do
+createCheckBox
+  :: QVBoxLayout.QVBoxLayout
+  -> CheckboxPosition
+  -> Text
+  -> IO QCheckBox.QCheckBox
+createCheckBox layout checkBoxPos labelText = do
   checkBox <- QCheckBox.new
-  QWidget.setSizePolicyRaw checkBox Maximum Maximum
+  QWidget.setSizePolicyRaw checkBox Maximum Preferred
```
maksbotan: Can you justify this change? AFAIR, i set `Maximum` so that checkbox never expands *its own label*, i.e. always remains a square.

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
```diff
@@ -69,7 +73,7 @@ data UiFeatures = UiFeatures
 ----------------------------------------------------------------------------
 
 -- API for the UI.
-data UiFace = UiFace
+newtype UiFace = UiFace
```
maksbotan: Again stale hlint fix.

pasqu4le:
@maksbotan I'll make this single comment to address all your required changes.

The password field of `UiRestoreWallet` is not wasted: see the `extractPass` function on line 98 of `ui/qt-app/Glue.hs` (used in line 88 of the same file).
In cases like this, where an UI has a password field, this is what happens: the password gets saved using the `WalletIdTemporary`, so the backend can have it without making another request to the UI. Being temporary this also means that it gets deleted as soon as this happens, meaning that if the user does another operation right after, he will need to input the password again. I just tried and I can confirm this.

I don't remember exactly the details about `Maximum` dropped in favor of `Preferred` except that I merged different functions in there and it did not make a difference for the UI elements that were there before, but did in those that I merged. Or something like that. I can look into it if you want.

The remaining changes are about those `newtype`s and missing new line, probably caused by all the rebasing, will fix them right now.

maksbotan:
> @maksbotan I'll make this single comment to address all your required changes.
> 
> The password field of `UiRestoreWallet` is not wasted: see the `extractPass` function on line 98 of `ui/qt-app/Glue.hs` (used in line 88 of the same file).
> In cases like this, where an UI has a password field, this is what happens: the password gets saved using the `WalletIdTemporary`, so the backend can have it without making another request to the UI. Being temporary this also means that it gets deleted as soon as this happens, meaning that if the user does another operation right after, he will need to input the password again. I just tried and I can confirm this.

Ooops, I did not pay enough attention, sorry. It's okay with me then. What happens if I want to create a wallet without password, i.e. don't check the checkbox?

> I don't remember exactly the details about `Maximum` dropped in favor of `Preferred` except that I merged different functions in there and it did not make a difference for the UI elements that were there before, but did in those that I merged. Or something like that. I can look into it if you want.

Please leave it as `Preferred`, but check that it looks OK just in case.

> The remaining changes are about those `newtype`s and missing new line, probably caused by all the rebasing, will fix them right now.

Okay, will accept after that.

pasqu4le:
> What happens if I want to create a wallet without password, i.e. don't check the checkbox?

Right now it's equivalent to creating a wallet with a password that is empty (`"" :: Text`), but AFAIK the discussion on whether it should be treated as a different thing is still open (using `Maybe Text` and `emptyPassphrase`) and I can easily change that again.

> check that it looks OK just in case

Just did and it seems to be looking fine. Fixed the other things as well.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Form/Send.hs
```diff
@@ -80,7 +80,7 @@ initSendWidget langFace walletIdxGetter accountsGetter =
     addWidgetChild WidgetNameSendAdd $
       initButtonWidget "+"
     addWidgetChild WidgetNameSendPass $
-      initPasswordWidget $ widgetParentLens sendPassL
+       initPasswordWidget $ widgetParentLens sendPassL
```
goodsoft: Space

gromakovsky:
Tested exceptions stuff in ghci, now everything works as expected.

ariadne/core/src/Ariadne/UX/PasswordManager.hs
```diff
@@ -41,31 +41,31 @@ data PasswordManager = PasswordManager
 
 createPasswordManager :: ComponentM PasswordManager
 createPasswordManager = buildComponent_ "Password Manager" $ do
-    passMapVar <- newIORef (Map.empty :: Map.Map WalletId (T.Text, UTCTime))
+    passMapVar <- newIORef (Map.empty :: Map.Map WalletId (T.Text, Time Second))
 
-    let nominalWait = 300 :: NominalDiffTime -- 5 minutes
+    let passLifeTime = minute 5
+        getCurrentSecs = sec . fromInteger . floor <$> getPOSIXTime
```
gromakovsky: Why not `realToFrac <$> getPOSIXTime`? Seems to be simpler and more precise.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -221,15 +221,18 @@ newAddress ::
        PassiveWalletLayer IO
     -> WalletFace
     -> IORef (Maybe WalletSelection)
+    -> (WalletReference -> IO PassPhrase)
+    -> (WalletReference -> IO HdAddress -> IO HdAddress)
+    -> WalletReference
```
gromakovsky: Why did this argument become needed here? `AccountReference` is already passed here and AFAIU it can be converted to `WalletReference`.

gromakovsky:
> Well, this looks like user-visible breakage. Don't know if it's severe though. @gromakovsky ?

Found this comment. There were many user-visible breakages since latest release (which was long ago), so I think breakages are not a problem now.

ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
pasqu4le: Did not know `WalletReference` could always be obtained from an `AccountReference`, removed it from the function and also reverted some code in `Knit` that was changed for this same reason

pasqu4le:
Yes `WalletIdSelected` is broken in that sense.

I would agree with resolving the selection to a `HdRootId` but it has 2 problem, that I was trying to solve to have a representation of `WalletRefByHdRootId` in `WalletId` before you told me that was already a problem for [AD-294]:
1. using a type variable was discussed before and it is not a good idea, because this would "propagate" a lot in the UIs, to the point that each `UiEvent` would need
2. assuming that the selection can be resolved in an `HdRootId` (I don't know how to yet), can we transform this in an `Text` (or something similar) that is unique before asking the password manager? If we can there should be no problem (`WalletId`s are always generated from `WalletReference`s, but the opposite never happens) and I think in that case I could use this to do that `WalletRefByHdRootId` representation I was mentioning before as well.

Another thing we could do that was not discussed (I had this idea, but I don't remember if I proposed it before) is to void the password for `WalletIdSelected` every time the selection changes. I know it's not ideal, but it's simple to do and at least the password manager would be aware of it and ask for a password again.

Regarding `WalletIdTemporary`: it may not seem very robust but as you said works in practice. Right now it should be impossible to break the "flow" that allows it to work as if it was atomical.

gromakovsky:
> after resolving a selection to a HdRootId, can we transform this in a Text (or something similar) that is unique before asking the password manager? 

Yes, I think we can and that's one of the options I proposed in my message.

> Another thing we could do that was not discussed (I had this idea, but I don't remember if I proposed it before) is to void the password for WalletIdSelected every time the selection changes. I know it's not ideal, but it's simple to do and at least the password manager would be aware of it and ask for a password again.

It seems simple indeed. Maybe let's do it in this PR and merge it, then reiterate in scope of AD-294?

pasqu4le:
> It seems simple indeed. Maybe let's do it in this PR and merge it, then reiterate in scope of AD-294?

Agreed, I'll do this.

>Yes, I think we can and that's one of the options I proposed in my message.

yep, sorry, I was trying to put the focus on the "unique" part, but I guess that's not a problem either.

gromakovsky:
> yep, sorry, I was trying to put the focus on the "unique" part, but I guess that's not a problem either.

Yes, not a problem, because wallet id is hash of it's public key, so we can use textual representation of this hash as `Text` identifier.

PR #313, opened 2018-09-15
Use `NonEmpty Word` as account path in GUI

maksbotan:
**Checklist:**

  - [x] [README](README.md)

**Description:**
Account paths can never actually be empty, so it's better to encode that
in types than constantly do matching and use `error`.

This makes AD-278 and AD-279 more easy to do.

This is a bit disputable, see line comments.


ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
```diff
@@ -176,6 +176,8 @@ type UiWalletTree = Tree UiWalletTreeItem
 -- to be an issue, we may consider changing it.
 type TreePath = [Word]
 
+type NonEmptyPath = NonEmpty Word
```
maksbotan: `TreePath` is also used for `wtsPath` and that probably can be empty.

`uadiPath` in `UiAddressInfo` should be non-empty too, but in fact it's requirement is stronger -- it should always be `[a, b]` -- account and address.

So maybe instead of storing this paths we store explicit indices? One for account and two for address.

gromakovsky: Probably the fact that `uadiPath` has exactly 2 indices is currency-specific and in other currencies it can be different, so I'd make it `NonEmptyPath`.

maksbotan: Okay. @goodsoft your word?

goodsoft: Won't break Disciplina, thus fine for now.

gromakovsky: So let's make it `NonEmptyPath` and I'll approve.

PR #315, opened 2018-09-17
[AD-377] mute read-only knit commands

pasqu4le:
**YT issue:** https://issues.serokell.io/issue/AD-377

**Checklist:**

- [x] Updated docs if necessary (Not necessary)

**Description:**

Remove read-only commands (`select`) in TUI from the `Repl` and the history, when launched by the UI.

### DIFF
--- a/ui/vty-app/Glue.hs
+++ b/ui/vty-app/Glue.hs
@@ -63,8 +63,9 @@ knitFaceToUI
   -> UiLangFace
 knitFaceToUI UiFace{..} KnitFace{..} =
   UiLangFace
-    { langPutCommand = putCommand Nothing
-    , langPutUiCommand = putUiCommand
+    { langPutCommand = putCommand False Nothing
+    , langPutUiCommand = putUiCommand False
+    , langPutUISilentCommand = putUiCommand True
     , langParse = Knit.parse
     , langPpExpr = Knit.ppExpr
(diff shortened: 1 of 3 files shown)


goodsoft:
I would rather make it responsibility of the call-site to hide the command.
That is, the widget itself would either call existing `putUiCommand` or another, (naming awful) `putUiROCommand` without echoing to Repl.

pasqu4le:
Agreed, would `putUISilentCommand` or `putUICommandMuted` be better?

Also because now it's only about read-only, but could be about other "silenced" commands in the future.

goodsoft:
The former looks better to me

pasqu4le:
Done & rebased

ui/vty-app/Glue.hs
```diff
@@ -63,31 +63,34 @@ knitFaceToUI
   -> UiLangFace
 knitFaceToUI UiFace{..} KnitFace{..} =
   UiLangFace
-    { langPutCommand = putCommand Nothing
-    , langPutUiCommand = putUiCommand
+    { langPutCommand = putCommand False Nothing
+    , langPutUiCommand = putUiCommand False
+    , langPutUISilentCommand = putUiCommand True
```
gromakovsky: A bit strange to see `xUIXx` among `xUiXx` (i. e. capital `I`), but since it's already merged, nevermind.

PR #319, opened 2018-09-21
[AD-279] Add Send dialog

maksbotan:
**Description:** A dialog to send transaction. There are some questions to discuss.

qtah commit was updated to include QEvent::Type PlatformSurface value.
Without this generic QEvent handler fails with `toEnum` error.

**YT issue:** https://issues.serokell.io/issue/AD-279

**Checklist:**

  - [x] [README](README.md)

### DIFF
--- a/ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
+++ b/ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
@@ -30,6 +30,7 @@ module Ariadne.UI.Qt.Face
 
 import qualified Control.Concurrent.Event as CE
 import Data.Loc.Span (Span)
+import Data.Scientific (Scientific)
 import Data.Tree (Tree)
 import Text.PrettyPrint.ANSI.Leijen (Doc)
 
@@ -91,7 +92,7 @@ data UiEvent
 -- | Commands issued by the UI widgets
 data UiCommand
   = UiSelect [Word]
-  | UiSend Text Text  -- ^ Address, amount
+  | UiSend Word [Word] Text Scientific -- ^ Wallet idx, accounts, address, amount
   | UiRestoreWallet Text (Maybe Text) Text Bool -- ^ Name, password, mnemonic, full restore
   | UiNewAccount Text  -- ^ Name
   | UiNewAddress Word Word -- ^ Wallet index, account index
@@ -142,6 +143,8 @@ data UiWalletFace =
   UiWalletFace
     { uiGenerateMnemonic :: Byte -> IO [Text]
     , uiDefaultEntropySize :: Byte
+    , uiValidateAddress :: Text -> Bool
+    , uiCoinPrecision :: Int
     }
 
 -- Interface for the command history
--- a/ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
+++ b/ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
@@ -1,0 +1,271 @@
+module Ariadne.UI.Qt.Widgets.Dialogs.Send
+  ( SendInputs(..)
+  , SendOptions(..)
+  , SendResult(..)
+  , runSend
+  ) where
+
+import Data.Scientific (Scientific, normalize, isInteger)
+import qualified Data.Text as T
+
+import Graphics.UI.Qtah.Signal (connect_)
+
+import Graphics.UI.Qtah.Core.HPoint (HPoint(..))
+import Graphics.UI.Qtah.Core.HSize (HSize(..))
+import qualified Graphics.UI.Qtah.Core.QEvent as QEvent
+import qualified Graphics.UI.Qtah.Core.QObject as QObject
+import qualified Graphics.UI.Qtah.Core.QLocale as QLocale
+import Graphics.UI.Qtah.Core.Types (QtWindowType(Popup))
+import qualified Graphics.UI.Qtah.Event as Event
+import qualified Graphics.UI.Qtah.Gui.QValidator as QValidator
+import qualified Graphics.UI.Qtah.Gui.QDoubleValidator as QDoubleValidator
+import qualified Graphics.UI.Qtah.Gui.QMouseEvent as QMouseEvent
+import qualified Graphics.UI.Qtah.Widgets.QAbstractButton as QAbstractButton
+import qualified Graphics.UI.Qtah.Widgets.QBoxLayout as QBoxLayout
+import qualified Graphics.UI.Qtah.Widgets.QCheckBox as QCheckBox
+import qualified Graphics.UI.Qtah.Widgets.QDialog as QDialog
+import qualified Graphics.UI.Qtah.Widgets.QHBoxLayout as QHBoxLayout
+import qualified Graphics.UI.Qtah.Widgets.QLabel as QLabel
+import qualified Graphics.UI.Qtah.Widgets.QLineEdit as QLineEdit
+import qualified Graphics.UI.Qtah.Widgets.QPushButton as QPushButton
+import qualified Graphics.UI.Qtah.Widgets.QVBoxLayout as QVBoxLayout
+import qualified Graphics.UI.Qtah.Widgets.QWidget as QWidget
+
+import Ariadne.UI.Qt.Face
+import Ariadne.UI.Qt.UI
+import Ariadne.UI.Qt.Util
+import Ariadne.UI.Qt.Widgets.Dialogs.Util
+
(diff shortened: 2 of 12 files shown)


ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
```diff
@@ -0,0 +1,252 @@
...
+initSend :: SendInputs -> IO Send
+initSend sendInputs = do
+  send <- QDialog.new
+  layout <- createLayout send
+
+  QWidget.setWindowTitle send ("SEND" :: String)
+
+  amountLabel <- QLabel.newWithText ("<b>AMOUNT</b>" :: String)
+  amountLayout <- QHBoxLayout.new
+  amountEdit <- QLineEdit.new
```
maksbotan: I can't use `QDoubleSpinBox` since it leads to one of two possible problems:

- If I get the value with `QDoubleSpinBox.value`, it's a `Double` and conversion to Scientific is unsafe theoretically
- If I get the value with `QDoubleSpinBox.text`, it's a string that depends on user's locale -- for me it has `,` as decimal separator, and we can't feed that string into `readMaybe` / `readEither`.

This however means that we must validate the value ourselves. And AFAIU we can't use `QDoubleValidator` for the same reason above and we can't subclass `QValidator` to make our own one.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
```diff
@@ -0,0 +1,252 @@
...
+    QDialog.Accepted -> case options of
+      Just so -> SendSuccess so
+      Nothing -> SendCancel
+    QDialog.Rejected -> SendCancel
+
+fillSendOptions :: Send -> IO (Maybe SendOptions)
+fillSendOptions Send{..} = do
+  amount <- fromString <$> QLineEdit.text amountEdit
+  address <- fromString <$> QLineEdit.text addressEdit
+  -- TODO do we want to validate any of this here? ^
```
maksbotan: I'd prefer to validate address format here, but I'm not sure how to do it properly.
Also, amount should probably be validated, since Cardano has precision only up to 10^-6, for example.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
```diff
@@ -268,8 +240,21 @@ requestButtonClicked langFace WalletInfo{..} _checked = do
...
+sendButtonClicked UiLangFace{..} WalletInfo{..} _checked = whenJustM (readIORef currentItem) $ \item -> do
+  let
+    (wIdx, inputs) = case item of
+      WIWallet wIdx uacis -> (wIdx, SendInputsMulti uacis)
+      WIAccount wIdx UiAccountInfo{..} -> (wIdx, SendInputsSingle $ head uaciPath)
+  result <- runSend inputs
+  case result of
+    SendCancel -> return ()
+    SendSuccess SendOptions{..} -> langPutUiCommand (UiSend wIdx soAccounts soAddress soAmount) >>= \case
+      Left err -> void $ QMessageBox.critical walletInfo ("Error" :: String) $ toString err
```
maksbotan: I'd display some kind of progress spinner here, need to discuss with @a-rukin. It takes some time between the dialog closes and backend does something with our transaction.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
gromakovsky: > I'd prefer to validate address format here, but I'm not sure how to do it properly.

Do you mean "which Cardano function to call" or "how to expose this function to UI"? In the former case it's `decodeTextAddress`, in the latter it should be exposed via knit somehow or, it appears to be too cumbersome, at least similar to `generateMnemonic`.

> Also, amount should probably be validated, since Cardano has precision only up to 10^-6, for example.

There is `Pos.Core.Common.Coin.checkCoin`, but it works with integer. Backend doesn't know about 10⁻⁶ precision, it should be specified somewhere in Glue.

ui/qt-lib/src/Ariadne/UI/Qt/Face.hs
```diff
@@ -28,6 +28,7 @@ module Ariadne.UI.Qt.Face
        ) where
 
 import Data.Loc.Span (Span)
+import Data.Scientific
```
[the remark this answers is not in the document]
maksbotan: :snake: :ok:

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
maksbotan: I remembered why I did not want to make `generateMnemonic` work through knit -- it disrupts the workflow: i have to send Knit comand and wait for the result _asynchronously_. That means that validation is no longer simple `Send -> IO Bool` function, it gets scattered into two places at least, and some state has to be stored somewhere.

I had an idea of making `KnitT` transformer that will hide this in its `>>=`, but probably nobody got time for that ATM.

So I'd go with `generateMnemonic`-like solution.

This has to be in Glue anyway.

CC @goodsoft.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
gromakovsky: Have you discussed it? Is it for later (some subsequent issue)?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
gromakovsky: > And AFAIU we can't use QDoubleValidator for the same reason

Wait, why not? AFAIU, `QDoubleValidator` is not really about `double` type.

> QDoubleValidator provides an upper bound, a lower bound, and a limit on the number of digits after the decimal point.

It doesn't force us to get `Double`, does it?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
gromakovsky: This comment is now obsolete, isn't it?

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
maksbotan: It doesn't. But it's still affected by locale, so will accept only commas. I can double check, but I'm pretty sure that's how it works.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
maksbotan: Right, will drop

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
maksbotan: Kostylee
Anyway, I fear that is inconvenient for the user too. I, for example, find this locale setting stupid and would be irritated by the need to enter the comma. And would not understand what's going on when I can enter dot.

maksbotan: What we __can__ do however, is to create the validator with `C` locale it whatever. What do you think?

gromakovsky: Is there some method to set locale for a validator? Sounds good to me then.

maksbotan: I had to add locales to qtah. Now it seems to work as intended.
Also please note that I had to disable group separators for locale. `C` locale has `,` as group separator, but `read` won't parse it.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/WalletInfo.hs
maksbotan: Updated AD-389.

ui/qt-lib/src/Ariadne/UI/Qt/Widgets/Dialogs/Send.hs
gromakovsky: What is group separator?

PR #321, opened 2018-09-21
[AD-167] Port wallet kernel tests

Pastafarianist:
**Description:**

This PR imports wallet kernel tests as of `cardano-sl` commit [`0fa0ce646cb83af8028878a1eed169f513a987b6`](https://github.com/input-output-hk/cardano-sl/tree/0fa0ce646cb83af8028878a1eed169f513a987b6).

Two tests do not pass yet. One of those failures is fixed in #328. The other is the subject of AD-416.

The failures can be seen by running the following command (warning: `--trace` turns on all stacktraces, which makes Stack rebuild 300+ packages):

```
stack test ariadne-cardano --trace --fast --test-arguments '--match "Compare wallet kernel to pure model"'
```

The difference to upstream can be seen via the following command (feel free to replace `kdiff3` with your favourite diff tool):

```
git difftool --tool=kdiff3 --dir-diff origin/iohk-wallet-new..origin/thatguy/ad167-port-iohk-tests-0fa0ce6 ariadne/cardano/test/backend/
```

TODO:

- [x] Make code adhere to `hlint` style (following merge of #304)

Overview of changes to the code:

- Put back `Kernel.DB.Read` and `Kernel.Util.StrictStateT` as they are used by the tests. We may move them to the tests though as the kernel does not use them.
- Removed `_hdAddressIsUsed` from `HdAddress` as we do not intend to use it.
- Changed `createHdWallet` to not throw `DuplicatedWalletKey` as an IO exception as the tests depend on this.
- Slightly refactored component creation functions to allow for `bracket*` functions. The functions themselves are moved into `Test.Spec.Fixture`.
- Fixed `instance Arbitrary HdRootId` where 12 bytes of entropy were generated instead of 32.
- Fixed test keystore creation to work around `UserSecret`'s inability to overwrite an existing but invalid file.
- Added `type UtxoByAccount = Map HdAccountId PrefilteredUtxo`.
- Updated `import_wallet.sh` to support tests.

**YT issue:** https://issues.serokell.io/issue/AD-167

**Checklist:**

- [x] Tested my changes if they modify code (pending)


ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -67,43 +67,54 @@ mkHasPP pp = HasNonemptyPassphrase $ pp /= emptyPassphrase
...
+                walletName
+                assuranceLevel
+                esk
+                utxoByAccount
+            case res2 of
+                Left e -> do
+                    liftIO $ Keystore.delete walletId (pw ^. walletKeystore)
+                    pure $ Left $ CreateWalletFailed e
+                Right hdRoot -> do
+                    pure $ Right hdRoot
```
gromakovsky: Is the only change here that we catch `DuplicatedWalletKey` and turn it into `CreateWalletFailed`?

ariadne/cardano/src/Ariadne/Wallet/Cardano/WalletLayer.hs
```diff
@@ -1,12 +1,32 @@
...
+
+import Ariadne.Wallet.Cardano.Kernel (PassiveWallet)
+import Ariadne.Wallet.Cardano.Kernel.Keystore (Keystore)
 import Ariadne.Wallet.Cardano.WalletLayer.Kernel
-  (passiveWalletLayerComponent, passiveWalletLayerWithDBComponent)
+  (passiveWalletLayerComponent, passiveWalletLayerInMemoryDBComponent)
+import Ariadne.Wallet.Cardano.WalletLayer.Types (PassiveWalletLayer)
 import Ariadne.Wallet.Cardano.WalletLayer.Types as Types
+
+bracketKernelPassiveWallet
```
gromakovsky: Similar comment as above. I think this function should be moved to tests or its name should reflect that it uses in-memory DB.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Read.hs
```diff
@@ -0,0 +1,88 @@
+module Ariadne.Wallet.Cardano.Kernel.DB.Read (
```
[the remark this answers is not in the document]
gromakovsky: So what about ↑?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -152,11 +154,19 @@ bracketTestKeystore withKeystore =
...
     Keystore <$> newMVar (InternalStorage us)
+  where
+    getKeystorePath :: IO FilePath
+    getKeystorePath = do
+        tempDir   <- getTemporaryDirectory
+        let dir = tempDir </> "ariadne-test"
+        createDirectoryIfMissing False dir
+        (fp, hdl) <- openTempFile dir "keystore.key"
+        hClose hdl
+        removeFile fp
```
gromakovsky: What's going on here? Why do we open a file, then close it and remove it?

Pastafarianist: The goal here is to get a path to the temp file. The file itself is created later. There is no simple API for path generation. The file cannot be left empty, because `UserSecret` cannot read empty files.

gromakovsky: Complicated, can you add a comment about that to this code please?

PR #327, opened 2018-09-28
[AD-296] add address & account for new wallets

epicallan:
**Description:**

**YT issue:** https://issues.serokell.io/issue/AD-

**Checklist:**

  - [ ] [README](README.md)
- [ ] My code complies with the [style guide](docs/code-style.md)


ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -35,22 +39,32 @@ import Ariadne.Wallet.Cardano.Kernel.Internal
...
 
 instance Arbitrary CreateWalletError where
     arbitrary = error "Arbitrary CreateWalletError is not implemented"
 
 instance Buildable CreateWalletError where
     build (CreateWalletFailed dbOperation) =
         bprint ("CreateWalletUnknownHdAccount " % F.build) dbOperation
+    build EmptyPassPhraseError = TB.fromText "EmptyPassPhraseError"
+    build (CreateAccountFailed dbOperation) =
+        bprint ("CreateWalletAcountUnknown " % F.build) dbOperation
```
Pastafarianist: Typo: `Acount`

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +200,25 @@ updateHdWalletAssurance
...
+    -> PassPhrase
+    -> HdRoot
+    -> IO (Either CreateWalletError HdRoot)
+addAddress pw walletId passphrase hdRoot = do 
+    accountRes <- createAccount (AccountName "un-named") walletId pw
+    case accountRes of 
+        Left err      -> return . Left $ CreateAccountFailed err
+        Right account -> do
+            let accountId = AccountIdHdSeq $ account ^. HD.hdAccountId
+            addressRes <- createAddress passphrase accountId HdChainInternal pw
```
Pastafarianist: `External` chain is [more suitable](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Change) here. Addresses on the internal chain should only ever be generated as change outputs.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +200,25 @@ updateHdWalletAssurance
...
+  Automatic creation of Wallet Account and Address
+-------------------------------------------------------------------------------}
+
+addAddress 
+    :: PassiveWallet
+    -> WalletId
+    -> PassPhrase
+    -> HdRoot
+    -> IO (Either CreateWalletError HdRoot)
+addAddress pw walletId passphrase hdRoot = do
```
Pastafarianist: Not sure why you need to take `HdRoot` as a parameter, only to return it in case of success. It would probably be better to return `IO (Either CreateWalletError ())` instead.

Also, you may want to use `ExceptT`. See examples in `Accounts` and `Addresses`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +200,25 @@ updateHdWalletAssurance
...
+-------------------------------------------------------------------------------}
+
+addAddress 
+    :: PassiveWallet
+    -> WalletId
+    -> PassPhrase
+    -> HdRoot
+    -> IO (Either CreateWalletError HdRoot)
+addAddress pw walletId passphrase hdRoot = do 
+    accountRes <- createAccount (AccountName "un-named") walletId pw
```
Pastafarianist: Please use something like `"Unnamed account"`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +200,25 @@ updateHdWalletAssurance
...
+    case accountRes of 
+        Left err      -> return . Left $ CreateAccountFailed err
+        Right account -> do
+            let accountId = AccountIdHdSeq $ account ^. HD.hdAccountId
+            addressRes <- createAddress passphrase accountId HdChainInternal pw
+            case addressRes of
+                Left err -> return . Left $ CreateAddressFailed err
+                Right _  -> return (Right hdRoot)
```
Pastafarianist: I'm also thinking that if creation of a new account/address fails, then this should be treated as a programmer error (this should never happen), and so all of these `Create{Account,Address}Failed` could be removed and the corresponding errors thrown with `impureThrow` according to the [exception guide](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/exceptions.md#potentially-impure-code-programmer-mistakes). However, this looks like a border case, and I'm OK with keeping all of these constructors too.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +200,25 @@ updateHdWalletAssurance
     -> IO (Either HD.UnknownHdRoot HdRoot)
 updateHdWalletAssurance pw hdRootId assuranceLevel =
     update' (pw ^. wallets) (UpdateHdWalletAssurance hdRootId assuranceLevel)
+
+{-------------------------------------------------------------------------------
+  Automatic creation of Wallet Account and Address
+-------------------------------------------------------------------------------}
+
+addAddress
```
Pastafarianist: Perhaps a better name would be something like `addDefaultAddress`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -175,3 +199,21 @@ updateHdWalletAssurance
...
+addDefaultAddress 
+    :: PassiveWallet
+    -> WalletId
+    -> PassPhrase
+    -> ExceptT CreateWalletError IO ()
+addDefaultAddress pw walletId passphrase = do 
+    res <- liftIO $ createAccount (AccountName "Unnamed account") walletId pw
+    case res of 
+        Left err      -> throwError $ CreateAccountFailed err
+        Right account -> do
```
Pastafarianist: This is better. You can also use `liftEither` like [here](https://github.com/serokell/ariadne/blob/624b379e508fcc926eca3dc8cf4c2af1d02a7147/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Accounts.hs#L106), but that's not essential.

epicallan: I don't think we can directly use liftEither as in the example you sited,  since `createAccount` returns  an `IO` with a different Exception type. I ended up refactoring it differently all the same.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Wallets.hs
```diff
@@ -35,22 +37,29 @@ import Ariadne.Wallet.Cardano.Kernel.Internal
...
-import Ariadne.Wallet.Cardano.Kernel.Types (WalletId(..))
+import Ariadne.Wallet.Cardano.Kernel.Types (WalletId(..), AccountId(..))
 import Ariadne.Wallet.Cardano.Kernel.Util (getCurrentTimestamp)
 
 import Test.QuickCheck (Arbitrary(..))
 
 
 data CreateWalletError =
-      CreateWalletFailed HD.CreateHdRootError
+      CreateWalletFailed HD.CreateHdRootError
```
Pastafarianist: Just one more minor thing. Please configure your editor to strip trailing whitespace.

PR #337, opened 2018-10-02
[AD-207] Add confirmation/cancel mechanism for important actions

pasqu4le:
**Description:** Add confirmation/cancel mechanism for important actions

**YT issue:** https://issues.serokell.io/issue/AD-207

**Checklist:**

- [X] Updated docs if necessary (unnecessary)
- [X] My code complies with the [style guide](docs/code-style.md)
- [X] Tested my changes if they modify code

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -303,14 +303,17 @@ newWallet ::
     -> WalletFace
     -> IO PassPhrase
     -> (ConfirmationType -> IO Bool)
+    -> Bool
     -> Maybe WalletName
     -> Maybe Byte
     -> IO [Text]
-newWallet pwl walletConfig face getPassTemp waitUiConfirm mbWalletName mbEntropySize = do
+newWallet pwl walletConfig face getPassTemp waitUiConfirm noConfirm mbWalletName mbEntropySize = do
   pp <- getPassTemp
   let entropySize = fromMaybe (wcEntropySize walletConfig) mbEntropySize
   mnemonic <- generateMnemonic entropySize
-  unlessM (waitUiConfirm $ ConfirmMnemonic mnemonic) $ throwM WGFailedUnconfirmedMnemonic
+  unless noConfirm $
+    unlessM (waitUiConfirm $ ConfirmMnemonic mnemonic) $
+      throwM WGFailedUnconfirmedMnemonic
   let seed = mnemonicToSeedNoPassword $ unwords $ Unsafe.init mnemonic
       (_, esk) = safeDeterministicKeyGen seed pp
   mnemonic <$
@@ -408,15 +411,18 @@ removeSelection
   -> WalletFace
   -> IORef (Maybe WalletSelection)
   -> (ConfirmationType -> IO Bool)
+  -> Bool
   -> IO ()
-removeSelection pwl WalletFace{..} walletSelRef waitUiConfirm = do
+removeSelection pwl WalletFace{..} walletSelRef waitUiConfirm noConfirm = do
   mWalletSel <- readIORef walletSelRef
   newSelection <- case mWalletSel of
     Nothing -> pure Nothing
     -- Throw "Nothing selected" here?
     Just selection -> do
       walletDb <- pwlGetDBSnapshot pwl
-      unlessM (waitUiConfirm $ ConfirmRemove walletDb selection) $ throwM RemoveFailedUnconfirmed
+      unless noConfirm $
+        unlessM (waitUiConfirm $ ConfirmRemove walletDb selection) $
+          throwM RemoveFailedUnconfirmed
       case selection of
         WSRoot hdrId -> do
           throwLeftIO $ void <$> pwlDeleteWallet pwl hdrId
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -310,8 +310,7 @@ newWallet pwl walletConfig face getPassTemp waitUiConfirm mbWalletName mbEntropy
   pp <- getPassTemp
   let entropySize = fromMaybe (wcEntropySize walletConfig) mbEntropySize
   mnemonic <- generateMnemonic entropySize
-  confirmed <- waitUiConfirm $ ConfirmMnemonic mnemonic
-  when (not confirmed) $ throwM WGFailedUnconfirmedMnemonic
+  unlessM (waitUiConfirm $ ConfirmMnemonic mnemonic) $ throwM WGFailedUnconfirmedMnemonic
   let seed = mnemonicToSeedNoPassword $ unwords $ Unsafe.init mnemonic
       (_, esk) = safeDeterministicKeyGen seed pp
   mnemonic <$
@@ -416,8 +415,7 @@ removeSelection pwl WalletFace{..} walletSelRef waitUiConfirm = do
     Nothing -> pure Nothing
     -- Throw "Nothing selected" here?
     Just selection -> do
-      confirmed <- waitUiConfirm $ ConfirmRemove selection
-      when (not confirmed) $ throwM RemoveFailedUnconfirmed
+      unlessM (waitUiConfirm $ ConfirmRemove selection) $ throwM RemoveFailedUnconfirmed
       case selection of
         WSRoot hdrId -> do
           throwLeftIO $ void <$> pwlDeleteWallet pwl hdrId
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
@@ -95,6 +95,13 @@ instance Exception SelectIsTooDeep where
   displayException SelectIsTooDeep =
     "The selection path is too deep"
 
+data WalletGenerationFailed = WGFailedUnconfirmedMnemonic
+  deriving (Eq, Show)
+
+instance Exception WalletGenerationFailed where
+  displayException WGFailedUnconfirmedMnemonic =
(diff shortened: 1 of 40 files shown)


ariadne/cardano/src/Ariadne/Wallet/Backend/KeyStorage.hs
```diff
@@ -289,13 +303,16 @@ newWallet ::
...
     -> Maybe WalletName
     -> Maybe Byte
     -> IO [Text]
-newWallet pwl walletConfig face getPassTemp mbWalletName mbEntropySize = do
+newWallet pwl walletConfig face getPassTemp waitUiConfirm mbWalletName mbEntropySize = do
   pp <- getPassTemp
   let entropySize = fromMaybe (wcEntropySize walletConfig) mbEntropySize
   mnemonic <- generateMnemonic entropySize
+  confirmed <- waitUiConfirm $ ConfirmMnemonic mnemonic
+  when (not confirmed) $ throwM WGFailedUnconfirmedMnemonic
```
maksbotan: `hlint` suggests `unless confirmed $` here. Actually you can use `unlessM (walletUiConfirm ...) $ throwM`  and avoid extra variable completely if you prefer so.

ui/vty-lib/src/Ariadne/UI/Vty/App.hs
```diff
@@ -369,5 +393,10 @@ handleAppWidgetEvent = \case
       UiPasswordRequest _ _ -> PasswordMode
       UiPasswordSent ->  NoModal
     resetAppFocus
+  UiConfirmEvent confirmEvent -> do
+    assignWidgetLens (Lens appModalL) $ case confirmEvent of
+      UiConfirmRequest _ confirmationType -> ConfirmationMode confirmationType
+      UiConfirmDone ->  NoModal
```
maksbotan: Two spaces after `->`?

ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -393,21 +414,27 @@ withWidgetState action = do
...
-drawWidgetChild :: WidgetName -> WidgetInfo s p -> WidgetNamePart -> B.Widget WidgetName
+drawWidgetChild :: WidgetName -> WidgetInfo s p -> WidgetNamePart -> WidgetDrawing
 drawWidgetChild focus widget@WidgetInfo{..} name =
-    maybe B.emptyWidget (drawWidget focus widget . withIgnoreVisibility) $ Map.lookup name widgetChildren
+    singleDrawing . maybe B.emptyWidget (last . drawWidget focus widget . withIgnoreVisibility) $ Map.lookup name widgetChildren
   where
     withIgnoreVisibility (Widget child) = Widget child{ widgetIgnoreVisibility = widgetIgnoreVisibility }
 
+singleDrawing :: (B.Widget WidgetName) -> WidgetDrawing
+singleDrawing = NE.fromList . (: [])
```
maksbotan: Isn't that just `(:| [])`? Also, redundant bracket in type signature.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmMnemonic.hs
```diff
@@ -0,0 +1,142 @@
...
+    ConfirmMnemonicWidgetState{..} <- get
+    whenJust confirmMnemonicWidgetResultVar $ \resultVar -> do 
+        case confirmMnemonicWidgetConfirmationState of
+            Before -> do
+                confirmMnemonicWidgetConfirmationStateL .= DisplayConfirmMnemonic
+            DisplayConfirmMnemonic -> do
+                confirmMnemonicWidgetConfirmationStateL .= RetypeConfirmMnemonic
+            RetypeConfirmMnemonic -> do
+                let mnemonic = confirmMnemonicWidgetValue
+                when (words (confirmMnemonicWidgetContent) == mnemonic) $ do
```
maksbotan: `()`?

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmMnemonic.hs
```diff
@@ -0,0 +1,142 @@
...
+                            [ B.txt "ConfirmMnemonic: "
+                            , last $ drawWidgetChild focus widget WidgetNameConfirmMnemonicInput
+                            ]
+                        ]
+
+handleConfirmMnemonicWidgetKey
+    :: KeyboardEvent
+    -> WidgetEventM ConfirmMnemonicWidgetState p WidgetEventResult
+handleConfirmMnemonicWidgetKey = \case
+    KeyEnter -> performContinue *> return WidgetEventHandled
```
maksbotan: Stupid nitpick `a *> return b` is just `a $> b`

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmRemove.hs
```diff
@@ -0,0 +1,92 @@
...
+        ]
+
+drawConfirmRemoveWidget
+    :: WidgetName
+    -> ConfirmRemoveWidgetState
+    -> WidgetDrawM ConfirmRemoveWidgetState p WidgetDrawing
+drawConfirmRemoveWidget focus ConfirmRemoveWidgetState{..} =
+    case confirmRemoveWidgetResultVar of
+        Nothing -> return $ singleDrawing B.emptyWidget
+        Just _  -> drawInsideDialog "Confirm ConfirmRemove" focus
```
maksbotan: Do you plan to add retyping to tui?

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmSend.hs
```diff
@@ -0,0 +1,94 @@
...
+
+
+drawConfirmSendWidget
+    :: WidgetName
+    -> ConfirmSendWidgetState
+    -> WidgetDrawM ConfirmSendWidgetState p WidgetDrawing
+drawConfirmSendWidget focus ConfirmSendWidgetState{..} = do
+    case confirmSendWidgetResultVar of
+        Nothing -> return $ singleDrawing B.emptyWidget
+        Just _  -> drawInsideDialog "Confirm ConfirmSend" focus [WidgetNameConfirmSendCancel, WidgetNameConfirmSendContinue] $ B.vBox $
```
maksbotan: What's up with these dialog names?

pasqu4le:
All the hlint issues / typos / wrong-dialog-names are there because I was in a hurry, I will fix them right now.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Dialog/ConfirmRemove.hs
pasqu4le: Yes, I will look putting that back. The fact is that the backend launches the event (that starts the dialog) with the data, and it doesn't have the name of the wallet/account.
This is why I removed it temporarily.

pasqu4le: Sorry I misread the question (I will leave the answer above as an answer to why retyping is not on GUI anymore). Anyway, I don't know, I suppose I will if I put it back on GUI.

maksbotan: You may want to take a look at what I did for new address command in gui. There I pass back account index from the original command, not from result. See resultToUI in glue.
If you get at least wallet index or root id back, you can look up its name in UI state, I suppose.

ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
gromakovsky: Yes, please avoid using `NE.fromList` wherever possible, because it's a partial function.

pasqu4le:
1. Very odd, I did not notice this, I will look into it right away
2/3. Sure thing

gromakovsky:
There's still one minor problem with TUI dialog: when I get to this stage https://i.imgur.com/C8ERBRa.png, `Continue` is selected (probably because it was selected right before I got there), it would be more convenient if line edit was selected automatically, because that's where I should type the mnemonic.

ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -393,21 +423,27 @@ withWidgetState action = do
...
 drawWidgetChild focus widget@WidgetInfo{..} name =
-    maybe B.emptyWidget (drawWidget focus widget . withIgnoreVisibility) $ Map.lookup name widgetChildren
+    singleDrawing . maybe B.emptyWidget (last . drawWidget focus widget . withIgnoreVisibility) $ Map.lookup name widgetChildren
   where
     withIgnoreVisibility (Widget child) = Widget child{ widgetIgnoreVisibility = widgetIgnoreVisibility }
 
+singleDrawing :: B.Widget WidgetName -> WidgetDrawing
+singleDrawing = (:| [])
+
+layeredDrawing :: [B.Widget WidgetName] -> WidgetDrawing
```
gromakovsky: I don't like this function, because it's partial. AFAIU, it has two goals:
1. Somewhat encapsulate type of `WidgetDrawing`, so that if it's changed, you'll probably be able to change only definition of this body so that it still accepts a list instead of modifying all places where it's called.
2. You don't need to import `Data.List.NonEmpty` where you want to do layered drawing.

I manually checked that in all places it's used with a fixed list consisting of two items, so all usages are safe.
However, to make it more explicit and checked by compiler I propose to change type of this function to `B.Widget WidgetName -> [B.Widget WidgetName] -> WidgetDrawing`. This way both goals will still be achieved and at the same time this function will never `error`.

pasqu4le: You are right about both goals, I will change the `layeredDrawing` function as you say.

PR #338, opened 2018-10-02
[AD-428] Always run UI action in the main thread

maksbotan:
**Description:** Always run UI action in the main thread. This is needed to fix Qt on macOS. Also, "Timers can only be stopped from the main thread" message is gone :tada: 

**YT issue:** https://issues.serokell.io/issue/AD-428

**Checklist:**

  - [x] [README](README.md)

### DIFF
--- a/ariadne/cardano/src/Ariadne/MainTemplate.hs
+++ b/ariadne/cardano/src/Ariadne/MainTemplate.hs
@@ -5,7 +5,7 @@ module Ariadne.MainTemplate
        , defaultMain
        ) where
 
-import Control.Concurrent.Async (race_, withAsync)
+import Control.Concurrent.Async (link, race_, withAsync)
 import Control.Monad.Component (ComponentM, runComponentM)
 import Control.Natural (($$))
 import Data.Version (Version)
@@ -154,7 +154,19 @@ initializeEverything MainSettings {..}
     serviceAction :: IO ()
     serviceAction =
       raceWithUpdateCheckAction $
-      uiAction `race_`
       cardanoAction
 
-  return $ withAsync initAction $ \_ -> serviceAction
+    mainAction :: IO ()
+    mainAction = do
+      initAction
+
+      -- Spawn backend actions in async thread, then run ui action in the main thread
+      -- This is needed because some UI libraries (Qt) insist on livng in the main thread
+      withAsync serviceAction $ \serviceThread -> do
+        -- Make backend rethrow all exceptions in main thread if something goes wrong
+        link serviceThread
+        -- TODO (AD-432) if uiAction blocks exceptions, for example by entering FFI call and not
+        -- returning from it, exceptions from the linked thread won't be rethrown correctly.
+        uiAction
+
+  return mainAction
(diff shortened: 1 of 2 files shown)


ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -154,7 +156,23 @@ initializeEverything MainSettings {..}
...
       cardanoAction
 
-  return $ withAsync initAction $ \_ -> serviceAction
+    mainAction :: IO ()
+    mainAction = do
+      -- First launch init action and wait until it completes
+      -- TODO is async really neded for that?
+      wait =<< async initAction
+      -- Then spawn backend actions in async thread
+      serviceThread <- async serviceAction
```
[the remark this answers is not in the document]
maksbotan: So the only reason we had the problem in the first place is that we run uiAction with race in serviceAction?

Do I understand right that you propose to run init and backend using withAsync and then run uiAction without any async in the current thread?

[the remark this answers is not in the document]
gromakovsky: > So the only reason we had the problem in the first place is that we run uiAction with race in serviceAction?

AFAIU, yes, because according to the docs it's equivalent to
```
race left right =
  withAsync left $ \a ->
  withAsync right $ \b ->
  waitEither a b
```

> Do I understand right that you propose to run init and backend using withAsync and then run uiAction without any async in the current thread?

I propose something like
```
  initAction
  withAsync serviceAction \serviceThread -> do
    link serviceThread
    uiAction
```

stack.yaml
```diff
@@ -21,7 +21,7 @@ packages:
 
 extra-deps:
 - git: https://github.com/serokell/cardano-sl
-  commit: a7c6f14c20df4b31acd3b552370185522f80c857 # ariadne branch
+  commit: 76e4e3f44d35c0139528cf11f284ff6c0cccf961 # ariadne branch
```
gromakovsky: It should be updated, I've just merged the `cardano-sl` PR.

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -154,7 +154,19 @@ initializeEverything MainSettings {..}
...
+    mainAction :: IO ()
+    mainAction = do
+      initAction
+
+      -- Spawn backend actions in async thread, then run ui action in the main thread
+      -- This is needed because some UI libraries (Qt) insist on livng in the main thread
+      withAsync serviceAction $ \serviceThread -> do
+        -- Make backend rethrow all exceptions in main thread if something goes wrong
+        link serviceThread
+        -- TODO if uiAction blocks exceptions, for example by entering FFI call and not returning from it,
```
gromakovsky: Mention AD-432 here please (AFAIU it's about AD-432).

ariadne/cardano/src/Ariadne/MainTemplate.hs
```diff
@@ -154,7 +154,19 @@ initializeEverything MainSettings {..}
...
+        -- exceptions from the linked thread won't be rethrown correctly.
+        uiAction
+
+  return $ mainAction
```
gromakovsky: `$` is redundant now, btw.

PR #342, opened 2018-10-08
[AD-53] Autocompletion

artemohanjanyan:
**Description:** Adds autocompletion.

##### Knit

This is where most of the changes are. Autocompletion is done via modifications of the parse tree of the command. I implemented "Trees that Grow" idiom in `Knit.Syntax` to make it possible to add extra information to the syntax tree.

`Knit.Parser` now returns `Expr ParseTreeExt` — expression with information about all tokens (tokens themselves and their spans). For autocompletion purposes, it can then be converted to `Expr FormattedExprExt` — expression with information about spacing between the nodes (the conversion is unique up to some cases with `OpUnit`). `Expr FormattedExprExt` then can be extended with completion suggestions and printed back to the user.

This approach has some drawbacks. User input should be correctly parsed in order to make completion possible. Now `Knit.Autocompletion` only appends missing closing parentheses, completion won't work for cases like `"new-"` (because it can't be tokenized) and `"kill id:"` (can't be parsed). Not sure what's the best way to work around these cases. I believe that ideally, we need to be able to parse only parts of the input.

##### TUI

In order to make autocompletion possible, I had to modify `Ariadne.UI.Vty.Widget.Form.Edit` so that I can change cursor position manually. Also, there is some code which calls autocompletion functions.

##### Other

I changed signatures of some functions exported from `Knit` and had to update their invocation in the rest of Ariadne.

Also, I wasn't sure how to format code in some places, looking forward to the comments.

**YT issue:** https://issues.serokell.io/issue/AD-53

**Checklist:**

  - [x] [README](README.md)


knit/src/Knit/Argument.hs
```diff
@@ -238,40 +247,42 @@ lookupArgSingle name args = do
...
 lookupArgOpt name = \case
     [] -> (Nothing, [])
-    ArgPos a : args -> (Just a, args)
-    arg@(ArgKw name' a) : args ->
+    ArgPos _ a : args -> (Just a, args)
+    arg@(ArgKw _ name' a) : args ->
         if name == name'
         then (Just a, args)
         else over _2 (arg:) $ lookupArgOpt name args
+    XArg _ : args -> lookupArgOpt name args
```
int-index: I see this handling of `XArg` in `lookupArgOpt` and `lookupArgMany`, but I'm not sure if it's a good idea.

The output of these functions should contain all unprocessed arguments, but you simply discard `XArg` without appending it to the list of processed arguments. If you want to simply ignore them, the correct treatment is this:

```
arg@(XArg _) : args -> over _2 (arg:) $ lookupArgOpt name args
```

However, I'm not sure that ignoring `XArg` is a good idea either. We probably would need to add a class for handling them. However, since it isn't needed for autocompletion, I propose one of two options:

1. Add a `XXArg ext ~ Void` constraint here and call `absurd`.
2. Remove the `XArg` constructor altogether.

knit/src/Knit/Argument.hs
```diff
@@ -286,17 +297,18 @@ consumeArguments ac args =
...
 toIrrelevanceError
-  :: [Arg (Value components)]
+  :: [Arg ext (Value components)]
   -> ArgumentError
 toIrrelevanceError = foldMap $ \case
-    ArgPos _ -> mempty { aeIrrelevantPos = 1 }
-    ArgKw key _ -> mempty { aeIrrelevantKeys = Set.singleton key }
+    ArgPos _ _ -> mempty { aeIrrelevantPos = 1 }
+    ArgKw _ key _ -> mempty { aeIrrelevantKeys = Set.singleton key }
+    XArg _ -> mempty
```
int-index: I don't think that would be a correct way to handle `XArg`. 

1. Add a `XXArg ext ~ Void` constraint here and call `absurd`.
2. Remove the `XArg` constructor altogether.

knit/src/Knit/Autocompletion.hs
```diff
@@ -0,0 +1,230 @@
...
+     )
+  => proxy components
+  -> Text
+  -> [Text]
+suggestions _ cmd =
+  let
+    openingParens = count "(" cmd
+    closingParens = count ")" cmd
+    addedParens = openingParens - closingParens + 1
+    cmd' = "(" <> cmd <> Data.Text.replicate addedParens ")"
```
int-index: That's a bad way to do this.

1. Assume the user's input contains a string literal with a parenthesis, e.g. `(foo x:")"`, you will detect as a closing parenthesis, but it isn't.
2. Printing the input string back will now contain parentheses that weren't there in the original user input (I see that you `drop` them a few lines below, but this relies on the fact that the pretty printer does not insert any whitespace between those parentheses - a fragile design).

I asked to modify the grammar to treat EOF as "close all parentheses", didn't this work? Why? Maybe ditch Earley and use parser combinators all the way, then it'll work?

artemohanjanyan: 1. My bad, indeed.
2. Example from 1 it should fail during parsing not proceed further. In general, I made another printer for `Expr FormattedTreeExt`, the only spaces it prints are spaces which are put in the tree explicitly. But I agree that that design is fragile.
3. I think that I misunderstood you, I didn't try that approach.

knit/src/Knit/Autocompletion.hs
```diff
@@ -0,0 +1,251 @@
...
+     , AllConstrained (ComponentTokenToLit components) components
+     , AllConstrained (ComponentCommandProcs components) components
+     , AllConstrained ComponentPrinter components
+     )
+  => proxy components
+  -> T.Text
+  -> [T.Text]
+suggestions _ cmd =
+  let
+    cmd' = "(" <> cmd
```
int-index: I'd like to see this parentheses business gone. Also, I don't like computations that involve spans.

Instead, let's change the tokenizer to return

```
newtype Skipped = Skipped String
(Skipped, [(SSpan, (Token components, Skipped))])
```

which is easy to achieve if we redefine `pSkip` as

```
pSkip = Skipped <$> many spaceChar
```

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Repl.hs
```diff
@@ -219,10 +235,50 @@ handleReplWidgetKey = \case
...
+        ls = T.splitOn "\n" replWidgetCommand
+        (locRow, locColumn) = replWidgetCommandLocation
+        cmdBeforeLocation = T.intercalate "\n"
+          $ snoc (take locRow ls) (T.take locColumn (ls !! locRow))
+        cmdAfterLocation = T.intercalate "\n"
+          $ T.drop locColumn (ls !! locRow) : drop (locRow + 1) ls
+        prefixSuggestions = langAutocomplete replWidgetLangFace cmdBeforeLocation
+        suggestions = (`map` prefixSuggestions) $ \suggestion ->
+          let ls' = T.splitOn "\n" suggestion
+          in (suggestion <> cmdAfterLocation, (length ls' - 1, length $ Universum.Unsafe.last ls'))
```
pasqu4le: You imported `Universum.Unsafe.last` and used it only here, but in other places (including one in `Ariadne.UI.Vty.Widget.Form.Edit`, and always to get the length of the last line) you use `T.length $ ls !! (length ls - 1)` that AFAIK has a higher cost (traverses the list twice: to get the length, and then to apply `!!`) and causes repetition.

Also you usually use `T.splitOn "\n"` and `T.intercalate "\n"` to divide and recombine lines, when there are functions in `Data.Text` to do exactly this.

What I suggest is:
- use `T.lines` instead of `T.splitOn "\n"`(NOTE: there is actually a difference between the two: `T.lines` will give back an empty list for a empty input, instead of a `[""]`, so you might need to change a bit of code to adjust the logic in `handleAutocompleteKey`)
- use `T.unlines` instead of `T.intercalate "\n"`
- write a function (for the `Edit` widget too) that calculates the position tuple out of a `Text`, that checks if it's empty (see 2 point above) and as such is in fact not unsafe anymore

I hope I explained myself well enough.

P.s. you can see from this very line that you can use `length` on `Text`, because `Universum`'s `Containers`' `length` has a `Data.Text` implementation (that uses `T.length`).

artemohanjanyan: `T.lines " \n"` also evaluates to `[" "]` while we need it to be `[" ",""]`. I suppose that that small difference between `T.lines` and `T.splitOn "\n"` was precisely the reason the latter was used in `Ariadne.UI.Vty.Widget.Form.Edit`, I just followed the example. Using `T.lines` and `T.unlines` will just add a lot of `if`s.

Agreed with all other points.

pasqu4le: Ok, if that's the case keep using them

artemohanjanyan: I think I'll leave position calculation in `Edit` as is because the scenario is slightly different there and just get rid of copy-paste in `Repl`.

PR #347, opened 2018-10-11
[AD-34] New keystore

epicallan:
**Description:**

**YT issue:** https://issues.serokell.io/issue/AD-34

**Checklist:**

  - [x] [README](README.md)

### DIFF
--- a/ariadne/cardano/src/Ariadne/Config/Wallet.hs
+++ b/ariadne/cardano/src/Ariadne/Config/Wallet.hs
@@ -23,7 +23,7 @@ import Ariadne.Util (postfixLFields)
 defaultWalletConfig :: FilePath -> WalletConfig
 defaultWalletConfig dataDir = WalletConfig
     { wcEntropySize = 16
-    , wcKeyfilePath = dataDir </> "secret-mainnet.key"
+    , wcKeyfilePath = dataDir </> "ariadne-secret.json"
     , wcAcidDBPath  = dataDir </> "wallet-db"
     }
 
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
@@ -1,14 +1,6 @@
 {-- | An opaque handle to a keystore, used to read and write 'EncryptedSecretKey'
       from/to disk.
-
-    NOTE: This module aims to provide a stable interface with a concrete
-    implementation concealed by the user of this module. The internal operations
-    are currently quite inefficient, as they have to work around the legacy
-    'UserSecret' storage.
-
 --}
-
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Ariadne.Wallet.Cardano.Kernel.Keystore
        ( Keystore -- opaque
@@ -17,7 +9,6 @@ module Ariadne.Wallet.Cardano.Kernel.Keystore
          -- * Constructing a keystore
        , keystoreComponent
        , bracketKeystore
-       , bracketLegacyKeystore
        -- * Inserting values
        , insert
        -- * Deleting values
@@ -35,39 +26,27 @@ import Prelude hiding (toList)
 import Control.Concurrent (modifyMVar_, withMVar)
 import Control.Monad.Component (ComponentM, buildComponent)
 import qualified Data.Map as Map
-import System.Directory
-  (createDirectoryIfMissing, getTemporaryDirectory, removeFile)
-import System.FilePath ((</>))
-import System.IO (hClose, openTempFile)
+import System.Directory (removeFile)
 
 import Pos.Core (AddressHash, addressHash)
 import Pos.Crypto (EncryptedSecretKey, PublicKey, encToPublic)
-import Pos.Util.UserSecret
-  (UserSecret, getUSPath, isEmptyUserSecret, peekUserSecret, usWallets,
-  writeUserSecret)
-import System.Wlog (CanLog(..), HasLoggerName(..), LoggerName(..), logMessage)
 
 import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet (HdRootId(..))
-import Ariadne.Wallet.Cardano.Kernel.DB.InDb (InDb(..), fromDb)
+import Ariadne.Wallet.Cardano.Kernel.DB.InDb (InDb(..))
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Persistence
+  (getTempKeystorePath, peekInternalStorage, writeInternalStorage)
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Types
+  (InternalStorage(..), isWallets)
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Util
+  (eskToPublicHash, walletIdToPublicHash)
 import Ariadne.Wallet.Cardano.Kernel.Types (WalletId(..))
 
--- Internal storage necessary to smooth out the legacy 'UserSecret' API.
-data InternalStorage = InternalStorage !UserSecret
+data Keystore
+  = Keystore !(MVar InternalStorage) !FilePath
 
--- A 'Keystore'.
-data Keystore = Keystore (MVar InternalStorage)
-
--- | Internal monad used to smooth out the 'WithLogger' dependency imposed
--- by 'Pos.Util.UserSecret', to not commit to any way of logging things just yet.
-newtype KeystoreM a = KeystoreM { fromKeystore :: IdentityT IO a }
-                    deriving (Functor, Applicative, Monad, MonadIO)
-
-instance HasLoggerName KeystoreM where
-    askLoggerName = return (LoggerName "Keystore")
-    modifyLoggerName _ action = action
-
-instance CanLog KeystoreM where
-    dispatchMessage _ln sev txt = logMessage sev txt
+{-------------------------------------------------------------------------------
+  Creating a keystore
+-------------------------------------------------------------------------------}
 
 -- | A 'DeletePolicy' is a preference the user can express on how to release
 -- the 'Keystore' during its teardown.
@@ -85,16 +64,6 @@ data DuplicatedWalletKey = DuplicatedWalletKey
 instance Exception DuplicatedWalletKey where
   displayException DuplicatedWalletKey =
     "The wallet with this root key already exists"
-
-{-------------------------------------------------------------------------------
-  Creating a keystore
--------------------------------------------------------------------------------}
-
--- FIXME [CBR-316] Due to the current, legacy 'InternalStorage' being
--- used, the 'Keystore' does not persist the in-memory content on disk after
--- every destructive operations, which means that in case of a node crash or
--- another catastrophic behaviour (e.g. power loss, etc) the in-memory content
--- not yet written in memory will be forever lost.
 
 keystoreComponent :: DeletePolicy -> FilePath -> ComponentM Keystore
 keystoreComponent deletePolicy fp =
@@ -117,25 +86,9 @@ bracketKeystore deletePolicy fp withKeystore =
 
 -- | Creates a new keystore.
 newKeystore :: FilePath -> IO Keystore
-newKeystore fp = runIdentityT $ fromKeystore $ do
-    us <- peekUserSecret fp
-    Keystore <$> newMVar (InternalStorage us)
-
--- | Creates a legacy 'Keystore' by reading the 'UserSecret' from a 'NodeContext'.
--- Hopefully this function will go in the near future.
-newLegacyKeystore :: UserSecret -> IO Keystore
-newLegacyKeystore us = Keystore <$> newMVar (InternalStorage us)
-
--- | Creates a legacy 'Keystore' using a 'bracket' pattern, where the
--- initalisation and teardown of the resource are wrapped in 'bracket'.
--- For a legacy 'Keystore' users do not get to specify a 'DeletePolicy', as
--- the release of the keystore is left for the node and the legacy code
--- themselves.
-bracketLegacyKeystore :: UserSecret -> (Keystore -> IO a) -> IO a
-bracketLegacyKeystore us withKeystore =
-    bracket (newLegacyKeystore us)
-            (\_ -> pass) -- Leave teardown to the legacy wallet
-            withKeystore
+newKeystore fp = do
+    istorage <- peekInternalStorage fp
+    Keystore <$> newMVar istorage <*> pure fp
 
 bracketTestKeystore :: (Keystore -> IO a) -> IO a
 bracketTestKeystore withKeystore =
@@ -153,47 +106,21 @@ bracketTestKeystore withKeystore =
 -- races due to the fact its underlying file is stored in the OS' temporary
 -- directory.
 newTestKeystore :: IO Keystore
-newTestKeystore = liftIO $ runIdentityT $ fromKeystore $ do
-    fp <- liftIO $ getKeystorePath
-    us <- peekUserSecret fp
-    Keystore <$> newMVar (InternalStorage us)
-  where
-    -- | Generates a new path for the temporary 'Keystore'.
-    -- Unfortunately, System.IO does not expose an API for path
-    -- generation (it is implemented in `openTempFile'`), so
-    -- we create the file, then close the handle and remove it.
-    -- It cannot be left empty since the `instance Bi` for
-    -- 'UserSecret' cannot handle empty files.
-    getKeystorePath :: IO FilePath
-    getKeystorePath = do
-        tempDir <- getTemporaryDirectory
-        let dir = tempDir </> "ariadne-test"
-        createDirectoryIfMissing False dir
-        (fp, hdl) <- openTempFile dir "keystore.key"
-        hClose hdl
-        removeFile fp
-        pure fp
+newTestKeystore = do
+    fp <- getTempKeystorePath
+    istorage <- peekInternalStorage fp
+    Keystore <$> newMVar istorage <*> pure fp
 
 -- | Release the resources associated with this 'Keystore'.
 releaseKeystore :: DeletePolicy -> Keystore -> IO ()
-releaseKeystore dp (Keystore ks) =
+releaseKeystore dp (Keystore mstorage fp) =
     -- We are not modifying the 'MVar' content, because this function is
     -- not exported and called exactly once from the bracket de-allocation.
-    withMVar ks $ \internalStorage@(InternalStorage us) -> do
-        fp <- release internalStorage
+    withMVar mstorage $ \istorage ->
         case dp of
              KeepKeystoreIfEmpty   -> pass
              RemoveKeystoreIfEmpty ->
-                 when (isEmptyUserSecret us) $ removeFile fp
-
--- | Releases the underlying 'InternalStorage' and returns the updated
--- 'InternalStorage' and the file on disk this storage lives in.
--- 'FilePath'.
-release :: InternalStorage -> IO FilePath
-release (InternalStorage us) = do
-    let fp = getUSPath us
-    writeUserSecret us
-    return fp
+                 when (isEmptyInternalStorage istorage) $ removeFile fp
 
 {-------------------------------------------------------------------------------
   Inserting things inside a keystore
@@ -204,16 +131,13 @@ insert :: WalletId
        -> EncryptedSecretKey
        -> Keystore
        -> IO ()
-insert _walletId esk (Keystore ks) = do
-    modifyMVar_ ks $ \(InternalStorage us) -> do
-        when (Map.member (eskToKey esk) (view usWallets us)) $
+insert _walletId esk (Keystore mstorage fp) = do
+    modifyMVar_ mstorage $ \istorage -> do
+        when (Map.member (eskToPublicHash esk) (view isWallets istorage)) $
             throwM DuplicatedWalletKey
-        let us' = us & over usWallets (Map.insert (addressHash $ encToPublic esk) esk)
-        writeUserSecret us'
-        return $ InternalStorage us'
-    where
-      eskToKey :: EncryptedSecretKey -> AddressHash PublicKey
-      eskToKey = addressHash . encToPublic
+        let istorage' = istorage & over isWallets (Map.insert (addressHash $ encToPublic esk) esk)
+        writeInternalStorage fp istorage'
+        return istorage'
 
 {-------------------------------------------------------------------------------
   Looking up things inside a keystore
@@ -223,13 +147,13 @@ insert _walletId esk (Keystore ks) = do
 lookup :: WalletId
        -> Keystore
        -> IO (Maybe EncryptedSecretKey)
-lookup wId (Keystore ks) =
-    withMVar ks $ \(InternalStorage us) -> return $ lookupKey us wId
+lookup wId (Keystore mstorage _) =
+    withMVar mstorage $ \istorage -> return $ lookupKey istorage wId
 
--- | Lookup a key directly inside the 'UserSecret'.
-lookupKey :: UserSecret -> WalletId -> Maybe EncryptedSecretKey
-lookupKey us walletId =
-    Map.lookup (walletIdToKey walletId) (us ^. usWallets)
+-- | Lookup a key directly inside the 'InternalStorage'.
+lookupKey :: InternalStorage -> WalletId -> Maybe EncryptedSecretKey
+lookupKey istorage walletId =
+    Map.lookup (walletIdToPublicHash walletId) (istorage ^. isWallets)
 
 {-------------------------------------------------------------------------------
   Deleting things from the keystore
@@ -238,11 +162,11 @@ lookupKey us walletId =
 -- | Deletes an element from the 'Keystore'. This is an idempotent operation
 -- as in case a key was not present, no error would be thrown.
 delete :: WalletId -> Keystore -> IO ()
-delete walletId (Keystore ks) = do
-    modifyMVar_ ks $ \(InternalStorage us) -> do
-        let us' = us & over usWallets (Map.delete (walletIdToKey walletId))
-        writeUserSecret us'
-        return (InternalStorage us')
+delete walletId (Keystore mstorage fp) = do
+    modifyMVar_ mstorage $ \istorage -> do
+        let istorage' = istorage & over isWallets (Map.delete (walletIdToPublicHash walletId))
+        writeInternalStorage fp istorage'
+        return istorage'
 
 {-------------------------------------------------------------------------------
   Converting a Keystore into container types
@@ -250,9 +174,9 @@ delete walletId (Keystore ks) = do
 
 -- | Returns all the 'EncryptedSecretKey' known to this 'Keystore'.
 toList :: Keystore -> IO [(WalletId, EncryptedSecretKey)]
-toList (Keystore ks) =
-    withMVar ks $ \(InternalStorage us) ->
-        pure $ map (first hashPubKeyToWalletId) $ toPairs $ us ^. usWallets
+toList (Keystore mstorage _) =
+    withMVar mstorage $ \istorage ->
+        pure $ map (first hashPubKeyToWalletId) $ toPairs $ istorage ^. isWallets
   where
     hashPubKeyToWalletId :: AddressHash PublicKey -> WalletId
     hashPubKeyToWalletId = WalletIdHdSeq . HdRootId . InDb
@@ -261,5 +185,6 @@ toList (Keystore ks) =
   Utilities
 -------------------------------------------------------------------------------}
 
-walletIdToKey :: WalletId -> AddressHash PublicKey
-walletIdToKey (WalletIdHdSeq hdRootId) = view fromDb $ getHdRootId hdRootId
+-- | Whether stored InternalStorage has empty keys
+isEmptyInternalStorage :: InternalStorage -> Bool
+isEmptyInternalStorage istorage = null (istorage ^. isWallets)
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
@@ -1,0 +1,105 @@
+module Ariadne.Wallet.Cardano.Kernel.Keystore.Types
+       ( StorageDecodingError (..)
+       , InternalStorage (..)
+       , Versioned (..)
+       , isWallets
+       , getInternalStorageKeys
+       ) where
+
+import Control.Lens.TH (makeLenses)
+import Data.Aeson
+  (FromJSON, ToJSON, Value(..), object, parseJSON, toJSON, withObject, (.:),
+  (.=))
+import Data.Aeson.Types (Parser, typeMismatch)
+import Data.Default (Default, def)
+import Data.Map (Map)
+import Formatting (bprint, build)
+
+import Pos.Binary.Class (decodeFull', serialize')
+import Pos.Core (AddressHash)
+import Pos.Crypto (EncryptedSecretKey(..), PublicKey)
+
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Util (esksToMap)
+
+import qualified Data.Map as Map (elems, empty)
+import qualified Data.Text.Buildable
+import qualified Data.Vector as V
+import qualified Serokell.Util.Base16 as S (decode, encode)
+
+-- | InternalStorage data containing wallet keys that gets saved to file
+data InternalStorage
+  = InternalStorage
+  { _isWallets  :: !(Map (AddressHash PublicKey) EncryptedSecretKey)
+  } deriving (Show)
+
+makeLenses ''InternalStorage
+
+instance Default InternalStorage where
+  def = InternalStorage { _isWallets = Map.empty }
+
+newtype Version = Version Int
+  deriving (Eq, Show, Generic)
+
+-- | Attaches a version to serialised JSON
+newtype Versioned a = Versioned a
+    deriving (Eq, Show)
+
+newtype StorageDecodingError = StorageDecodingError Text
+    deriving (Show)
+
+instance Exception StorageDecodingError
+
+instance Buildable StorageDecodingError where
+    build (StorageDecodingError msg) =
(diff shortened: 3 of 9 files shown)


epicallan:
We are unfortunately still using `Bi` instances due to the fact that we cannot have  aeson JSON instances for a some data types from `cardano` since they are in non exposed modules.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,92 @@
...
+                in UserSecret <$> pEncMap <*> pVersion <*> pure mempty
+
+{-------------------------------------------------------------------------------
+  Utilities
+-------------------------------------------------------------------------------}
+
+esksToMap :: [EncryptedSecretKey] -> Map (AddressHash PublicKey) EncryptedSecretKey
+esksToMap esks =
+    esks <&> (\encKey -> (eskToKey encKey, encKey))
+          & Map.fromList
```
Pastafarianist: How about

```haskell
esksToMap = Map.fromList . map (\encKey -> (eskToKey encKey, encKey))
```

?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,92 @@
...
+data UserSecret
+  = UserSecret
+  { _usWallets  :: Map (AddressHash PublicKey) EncryptedSecretKey
+  , _usVersion  :: !Version
+  , _usFilePath :: !FilePath
+  } deriving (Show)
+
+makeLenses ''UserSecret
+
+instance Default UserSecret where
```
Pastafarianist: Since this instance is _always_ used like so
```haskell
def & usFilePath .~ fp
```
perhaps it would be better to provide a separate function, like `makeUserSecret :: FilePath -> UserSecret` or something like that?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,92 @@
...
+       , getUserSecretKeys
+       , usFilePath
+       , usWallets
+       , usVersion
+       ) where
+
+import Control.Lens.TH (makeLenses)
+import Data.Aeson
+  (FromJSON, ToJSON, Value(..), toJSON, parseJSON, withText, withObject, (.:), object, (.=))
+import Data.Default
```
Pastafarianist: Please avoid wildcard imports.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
+-- | system file operations related to key-store UserSecret
```
Pastafarianist: Please use a capital letter at the beginning and a full stop at the end.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
...
+
+module Ariadne.Wallet.Cardano.Kernel.Keystore.Util
```
Pastafarianist: This module is related to I/O operations on the new `UserSecret`. Perhaps a better name for this module would be something like `Persistence` or `IO`?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
...
+-- | Creates empty temporary storage file for keyStore testing purposes
+getTempKeystorePath :: MonadIO m => m FilePath
+getTempKeystorePath = do
+    tempDir <- liftIO getTemporaryDirectory
+    let dir = tempDir </> "ariadne-test"
+    liftIO $ createDirectoryIfMissing False dir
+    liftIO $ createEmptyFile $ dir </> "keystore.json"
+
+createEmptyFile :: FilePath -> IO FilePath
+createEmptyFile fp = writeFile fp mempty $> fp
```
Pastafarianist: Why would you return the parameter from the function unchanged?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
...
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Types
+  (UserSecret(..), StorageDecodingError(..), usWallets, usFilePath)
+
+
+lockFilePath :: FilePath -> FilePath
+lockFilePath = (<> ".lock")
+
+-- | Creates empty temporary storage file for keyStore testing purposes
+getTempKeystorePath :: MonadIO m => m FilePath
+getTempKeystorePath = do
```
Pastafarianist: It is perhaps better to wrap the whole function in `liftIO $ do`.

ariadne/cardano/src/Ariadne/System/FileMode.hs
```diff
@@ -0,0 +1,39 @@
...
+    mode <- liftIO $ PSX.fileMode <$> PSX.getFileStatus path
+    return $ PSX.intersectFileModes mode PSX.accessModes
+
+-- | Set mode disregard current mode of the file.
+setMode :: MonadIO m => PSX.FileMode -> FilePath -> m ()
+setMode mode path = liftIO $ PSX.setFileMode path mode
+
+-- | Set given mode if needed.
+ensureModeIs
+    :: (MonadIO m, CanLog m)
```
Pastafarianist: Do I understand correctly that this function is missing the `IsPosix` constraint?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
...
+    -- hence no handler.
+    (tempPath, tempHandle) <-
+        openBinaryTempFile (takeDirectory path) (takeFileName path)
+
+    LBS.hPut tempHandle (encode us) `finally` hClose tempHandle
+    hClose tempHandle
+    renameFile tempPath path
+
+-- | Whether stored UserSecret has empty keys
+isEmptyUserSecret :: UserSecret -> Bool
```
Pastafarianist: This seems like the only function in this module that is not related to I/O. Furthermore, it is not used anywhere in this module. Perhaps it should be moved elsewhere?

ariadne/cardano/test/backend/Test/Spec/Keystore.hs
```diff
@@ -4,12 +4,12 @@ module Test.Spec.Keystore (
...
   ) where
 
-import System.Directory (doesFileExist, getTemporaryDirectory, removeFile)
-import System.FilePath ((</>))
+import System.Directory (doesFileExist, removeFile)
 import System.IO.Error (IOError)
 
 import Test.Hspec (Spec, describe, it, shouldBe, shouldReturn)
 import Test.Hspec.QuickCheck (prop)
+import Test.Pos.Core.Arbitrary ()
```
Pastafarianist: Why do you need this import?

ariadne/cardano/src/Ariadne/System/FileMode.hs
```diff
@@ -0,0 +1,39 @@
+module Ariadne.System.FileMode
+       ( getAccessMode
+       , setMode
+       , ensureModeIs
+       , mode600
```
Pastafarianist: Are you sure that all of these exports are needed? You may run `weeder` on the codebase to see which of them are never used outside of this module.

ariadne/cardano/src/Ariadne/System/FileMode.hs
```diff
@@ -0,0 +1,39 @@
...
+mode600 :: PSX.FileMode
+mode600 = PSX.unionFileModes PSX.ownerReadMode PSX.ownerWriteMode
+
+-- | Return only the access part of the file mode (like owner:rw-, etc).
+getAccessMode :: MonadIO m => FilePath -> m PSX.FileMode
+getAccessMode path = do
+    mode <- liftIO $ PSX.fileMode <$> PSX.getFileStatus path
+    return $ PSX.intersectFileModes mode PSX.accessModes
+
+-- | Set mode disregard current mode of the file.
```
Pastafarianist:
```suggestion
-- | Set mode regardless of the current mode of the file.
```

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,92 @@
...
+esksToMap :: [EncryptedSecretKey] -> Map (AddressHash PublicKey) EncryptedSecretKey
+esksToMap esks =
+    esks <&> (\encKey -> (eskToKey encKey, encKey))
+          & Map.fromList
+
+eskToKey :: EncryptedSecretKey -> AddressHash PublicKey
+eskToKey = addressHash . encToPublic
+
+getUserSecretKeys :: UserSecret -> [EncryptedSecretKey]
+getUserSecretKeys us = Map.elems (us ^. usWallets)
```
Pastafarianist: Actually, `eskToKey` and `getUserSecretKeys` look a lot like functions that could be put in a `Util` module.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -251,8 +206,8 @@ delete walletId (Keystore ks) = do
 -- | Returns all the 'EncryptedSecretKey' known to this 'Keystore'.
 toList :: Keystore -> IO [(WalletId, EncryptedSecretKey)]
 toList (Keystore ks) =
-    withMVar ks $ \(InternalStorage us) ->
-        pure $ map (first hashPubKeyToWalletId) $ toPairs $ us ^. usWallets
+    withMVar ks $ \us ->
+        pure $ map (first hashPubKeyToWalletId) $ Map.toList $ us ^. usWallets
```
Pastafarianist: Why replace `toPairs` here?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
epicallan: Damn mistake. 😞

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
epicallan: I agree. 👍

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,97 @@
...
+writeRaw path us = do
+    -- On POSIX platforms, openBinaryTempFile guarantees that the file
+    -- will be created with mode 600.
+    -- If openBinaryTempFile throws, we want to propagate this exception,
+    -- hence no handler.
+    (tempPath, tempHandle) <-
+        openBinaryTempFile (takeDirectory path) (takeFileName path)
+
+    LBS.hPut tempHandle (encode us) `finally` hClose tempHandle
+    hClose tempHandle
```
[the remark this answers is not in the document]
epicallan: Another Damn mistake. Shouldn't be there 😞.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
epicallan: Agreed!

ariadne/cardano/test/backend/Test/Spec/Keystore.hs
epicallan: I was getting errors related to lack of  class instances from `Test.Pos.Core.Arbitrary () `.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
epicallan: Good call

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
epicallan: I didn't realize then that they are essentially same thing. Will revert.

ariadne/cardano/src/Ariadne/Config/Wallet.hs
```diff
@@ -23,7 +23,7 @@ import Ariadne.Util (postfixLFields)
 defaultWalletConfig :: FilePath -> WalletConfig
 defaultWalletConfig dataDir = WalletConfig
     { wcEntropySize = 16
-    , wcKeyfilePath = dataDir </> "secret-mainnet.key"
+    , wcKeyfilePath = dataDir </> "secret-mainnet.json"
```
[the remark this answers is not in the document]
epicallan: I think its better.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,92 @@
...
+
+import qualified Data.Text.Buildable
+import qualified Data.Map as Map (empty, fromList, elems)
+import qualified Serokell.Util.Base16 as S (encode, decode)
+
+newtype Version= Version { getVersion :: Int }
+  deriving (Show, Eq)
+
+-- | UserSecret data containing wallet keys that gets saved to file
+data UserSecret
```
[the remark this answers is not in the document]
epicallan: Makes sense. Will switch to `InternalStorage`

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
epicallan: Turns we can't have `getUserSecretKeys` in a `Util` module due to `cyclic dependence issue` between the 2 modules. I added some other functions there though.

ariadne/cardano/src/Ariadne/System/Other.hs
```diff
@@ -0,0 +1,23 @@
...
+import System.Info (os)
+
+---------------------------------------------------------------------
+-- OS-dependent operations
+---------------------------------------------------------------------
+
+data PosixOperationsEnabled = PosixOperationsEnabled
+
+-- | Allowance for POSIX operations.
+type IsPosix = Given PosixOperationsEnabled
```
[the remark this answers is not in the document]
epicallan: Its if we care about not throwing IO exceptions in a `Non Posix OS`

ariadne/cardano/src/Ariadne/System/FileMode.hs
epicallan: I thought so to, but adding it there triggers `warning: [-Wredundant-constraints]`

ariadne/cardano/src/Ariadne/System/FileMode.hs
epicallan: removed the ones not used. The code snippet here hasn't updated to reflect that though.

ariadne/cardano/src/Ariadne/System/FileMode.hs
Pastafarianist: Please see the discussion in #haskell in Slack. The whole approach with `IsPosix` is not going to work. I suggest that you get rid of all mentions of it.

ariadne/cardano/src/Ariadne/System/Other.hs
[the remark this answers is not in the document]
Pastafarianist: Please see the discussion in #haskell in Slack. The whole approach with IsPosix is not going to work. I suggest that you get rid of all mentions of it.

ariadne/cardano/src/Ariadne/Wallet/Backend/Restore.hs
```diff
@@ -91,10 +91,7 @@ restoreFromKeyFile ::
     -> WalletRestoreType
     -> IO ()
 restoreFromKeyFile pwl face runCardanoMode mbWalletName path rType = do
-    keyFile <- BS.readFile path
-    us <- case decodeFull' keyFile of
-      Left e   -> throwM $ SecretsDecodingError path e
-      Right us -> pure us
+    us <- readInternalStorage path
```
Pastafarianist: `istorage`?

ariadne/cardano/test/backend/Test/Spec/Keystore.hs
Pastafarianist: Strange that they were not required before.

ariadne/cardano/src/Ariadne/System/FileMode.hs
Martoon-00: IIUC In that discussion in #haskell you tried to use `IsPosix` constraint to make sure that `whenPosix` _is not_ called when it _is not_ necessary. while its original purpose was to ensure `whenPosix` _is_ used when it _is_ necessary.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -218,3 +210,7 @@ toList (Keystore ks) =
 
 walletIdToKey :: WalletId -> AddressHash PublicKey
 walletIdToKey (WalletIdHdSeq hdRootId) = view fromDb $ getHdRootId hdRootId
+
+-- | Whether stored InternalStorage has empty keys
+isEmptyInternalStorage :: InternalStorage -> Bool
+isEmptyInternalStorage istorage = null (istorage ^. isWallets)
```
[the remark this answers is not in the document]
epicallan: Configured it right. Shouldn't happen again.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,95 @@
...
+    build (StorageDecodingError msg) =
+        "Failed to decode stored json: " <> bprint build msg
+
+instance ToJSON EncryptedSecretKey where
+  toJSON = String . S.encode . serialize'
+
+instance FromJSON EncryptedSecretKey where
+  parseJSON  = withText "EncryptedSecretKey" $ \encryptedText ->
+                let eitherDecode = either textFail pure . decodeFull'
+                in either textFail eitherDecode $ S.decode encryptedText
```
Pastafarianist: Please don't indent code like this. Add a linebreak after `parseJSON =`, then add 4 spaces per indentation level. Also please make sure you use 4 spaces everywhere.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,95 @@
...
+  toJSON istorage =
+    object [ "keys"    .= getInternalStorageKeys istorage
+           , "version" .= (toText $ showVersion $ istorage ^. isVersion)
+           ]
+
+instance FromJSON InternalStorage where
+  parseJSON = withObject "InternalStorage" $ \v ->
+                let pEncMap  = esksToMap <$> v .: "keys"
+                    pVersion = either textFail pure . decodeVersion =<< v .: "version"
+                in InternalStorage <$> pEncMap <*> pVersion <*> pure mempty
```
Pastafarianist: Same as above.

Pastafarianist:
Also, please ensure that `stylish-haskell` is applied to all affected files before merging.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,95 @@
...
+import  Ariadne.Wallet.Cardano.Kernel.Keystore.Util (esksToMap)
+
+import qualified Data.Text.Buildable
+import qualified Data.Map as Map (elems, empty)
+import qualified Serokell.Util.Base16 as S (encode, decode)
+
+-- | InternalStorage data containing wallet keys that gets saved to file
+data InternalStorage
+  = InternalStorage
+  { _isWallets  :: Map (AddressHash PublicKey) EncryptedSecretKey
```
Pastafarianist:
```suggestion
  { _isWallets  :: !(Map (AddressHash PublicKey) EncryptedSecretKey)
```

ariadne/cardano/test/backend/Test/Spec/Keystore.hs
```diff
@@ -70,11 +66,11 @@ spec =
                     (fmap hash mbKey) `shouldBe` (Just (hash esk))
 
         prop "Inserts are persisted after releasing the keystore" $ monadicIO $ do
-            keystorePath <- getKeystorePath
+            keystorePath <- getTempKeystorePath
             (STB wid, STB esk) <- pick genKeypair
             run $ do
                 nukeKeystore keystorePath
-                Keystore.bracketKeystore KeepKeystoreIfEmpty keystorePath $ \keystore1 ->
+                Keystore.bracketKeystore KeepKeystoreIfEmpty keystorePath $ \keystore1 -> do
```
gromakovsky: Redundant change.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
```diff
@@ -0,0 +1,15 @@
...
+
+import Pos.Core (AddressHash, addressHash)
+import Pos.Crypto (EncryptedSecretKey(..), PublicKey, encToPublic)
+
+import qualified Data.Map as Map (fromList)
+
+esksToMap :: [EncryptedSecretKey] -> Map (AddressHash PublicKey) EncryptedSecretKey
+esksToMap = Map.fromList . map (\encKey -> (eskToKey encKey, encKey))
+
+eskToKey :: EncryptedSecretKey -> AddressHash PublicKey
```
gromakovsky: Why is it called `eskToKey` if it returns hash?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,95 @@
...
+    deriving (Show)
+
+instance Exception StorageDecodingError
+
+instance Buildable StorageDecodingError where
+    build (StorageDecodingError msg) =
+        "Failed to decode stored json: " <> bprint build msg
+
+instance ToJSON EncryptedSecretKey where
+  toJSON = String . S.encode . serialize'
```
gromakovsky: Let's avoid orphan instances. This instance is not necessary.
You can define `eskToJSON = String . S.encode . serialize'` and apply `map eskToJSON` to the result of `getInternalStorageKeys`.
Similar thing can be done for `fromJSON`.

ariadne/cardano/src/Ariadne/System/FileMode.hs
```diff
@@ -0,0 +1,47 @@
...
+       ) where
+
+import System.Info (os)
+import System.Wlog (logWarning, usingLoggerName, CanLog)
+
+import qualified System.Posix.Files as PSX
+import qualified System.Posix.Types as PSX (FileMode)
+
+-- | It's common practice to make keyfiles readable/writable for user only.
+-- For now keyfile is read-only
```
gromakovsky: > For now keyfile is read-only

What does it mean?

ariadne/cardano/src/Ariadne/System/FileMode.hs
```diff
@@ -0,0 +1,47 @@
+module Ariadne.System.FileMode
```
gromakovsky: I think it's better to put this module into `ariadne-core`, because it doesn't depend on anything from Cardano.

ariadne/cardano/src/Ariadne/Wallet/Backend/Restore.hs
```diff
@@ -91,10 +91,7 @@ restoreFromKeyFile ::
     -> WalletRestoreType
     -> IO ()
 restoreFromKeyFile pwl face runCardanoMode mbWalletName path rType = do
-    keyFile <- BS.readFile path
-    us <- case decodeFull' keyFile of
-      Left e   -> throwM $ SecretsDecodingError path e
```
gromakovsky: AFAIU, after this deletion `SecretsDecodingError` (defined in this file) is not used anymore, so it should be deleted.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -117,23 +87,23 @@ bracketKeystore deletePolicy fp withKeystore =
...
 
 -- | Creates a legacy 'Keystore' using a 'bracket' pattern, where the
 -- initalisation and teardown of the resource are wrapped in 'bracket'.
 -- For a legacy 'Keystore' users do not get to specify a 'DeletePolicy', as
 -- the release of the keystore is left for the node and the legacy code
 -- themselves.
-bracketLegacyKeystore :: UserSecret -> (Keystore -> IO a) -> IO a
-bracketLegacyKeystore us withKeystore =
-    bracket (newLegacyKeystore us)
+bracketLegacyKeystore :: InternalStorage -> (Keystore -> IO a) -> IO a
```
gromakovsky: Let's get rid of legacy. That's the primary goal of the issue. Fortunately, this function seems to be unused, so it can be simply deleted.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -263,3 +210,7 @@ toList (Keystore ks) =
 
 walletIdToKey :: WalletId -> AddressHash PublicKey
 walletIdToKey (WalletIdHdSeq hdRootId) = view fromDb $ getHdRootId hdRootId
+
+-- | Whether stored InternalStorage has empty keys
+isEmptyInternalStorage :: InternalStorage -> Bool
```
gromakovsky: Tbh, I do not understand why `getInternalStorageKeys` is in `Types` and this function is here. Both are simple pure functions which take `InternalStorage` and return something, so I think they should be in the same module :man_shrugging:

ariadne/cardano/test/backend/Test/Spec/Keystore.hs
Pastafarianist: I think it's OK as it makes the code more uniform.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Util.hs
epicallan: I think I adopted from previous code, I have changed.

ariadne/cardano/src/Ariadne/System/FileMode.hs
epicallan: Corrected it to `-- For now keyfile is read and write only`

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
epicallan: The reason is to avoid cyclic dependencies. An excellent place to put `getInternalStorageKeys` would be `Ariadne.Wallet.Cardano.Kernel.Keystore.Util` but then it would lead to cyclic dependencies since it would require `InternalStorage` in `Ariadne.Wallet.Cardano.Kernel.Keystore.Types` . `isEmptyInternalStorage` would also have a similar issue

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -153,47 +106,27 @@ bracketTestKeystore withKeystore =
...
 
 -- | Releases the underlying 'InternalStorage' and returns the updated
 -- 'InternalStorage' and the file on disk this storage lives in.
--- 'FilePath'.
 release :: InternalStorage -> IO FilePath
-release (InternalStorage us) = do
-    let fp = getUSPath us
-    writeUserSecret us
-    return fp
+release istorage = writeInternalStorage istorage $> istorage  ^. isFilePath
```
gromakovsky: It seems to be the only place where `isFilePath` is actually used. And that can be easily avoided. `release` is only called from `releaseKeystore`. It's called from `keystoreComponent`, `bracketKeystore` and `bracketTestKeystore`. The first two can pass `FilePath` to `releaseKeystore` already. `bracketTestKeystore` needs to be changed a little bit so that `newTestKeystore` returns a pair.

I don't understand this design when `FilePath` is part of `Keystore`. I know it was like that before, but I'd like to get rid of that unless someone provides a good reason why it's good.

One reason why it's bad is that when we parse `JSON` we return `Keystore` with some static (empty) file path. That's already enough for me to dislike this design.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -263,3 +193,7 @@ toList (Keystore ks) =
 
 walletIdToKey :: WalletId -> AddressHash PublicKey
```
gromakovsky: Btw, I have a similar question to the one I had about `eskToKey`. It returns a hash, not a key.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,101 @@
...
+import qualified Data.Map as Map (elems, empty)
+import qualified Data.Text.Buildable
+import qualified Data.Vector as V
+import qualified Serokell.Util.Base16 as S (decode, encode)
+
+-- | InternalStorage data containing wallet keys that gets saved to file
+data InternalStorage
+  = InternalStorage
+  { _isWallets  :: !(Map (AddressHash PublicKey) EncryptedSecretKey)
+  , _isVersion  :: !Version
```
gromakovsky: We don't need this field in the data type. We don't use it anywhere. We should put `version` into serialized JSON value and check it during deserialization. But we shouldn't carry it further.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore/Types.hs
```diff
@@ -0,0 +1,101 @@
...
+
+eskToJSON :: EncryptedSecretKey -> Value
+eskToJSON = String . S.encode . serialize'
+
+eskFromJson :: Value -> Parser EncryptedSecretKey
+eskFromJson (String encryptedText) =
+    let eitherDecode = either textFail pure . decodeFull'
+    in either textFail eitherDecode $ S.decode encryptedText
+
+eskFromJson _ = fail "Parsing EncryptedSecretKey Array error"
```
gromakovsky: It's better to use `typeMismatch` from `aeson`.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
epicallan: @gromakovsky The reason why we have filePath as part of `keystore` is that it's passed as a configuration value from ` WalletConfig` in `wcKeyfilePath`

if we derive the filePath from within `Ariadne.Wallet.Cardano.Kernel.Keystore.Persistence` then we can get rid of filePath in `keystore`.

epicallan: @gromakovsky The reason is that it allows us to use keystore operations such as `insert` and `delete` without use of a `filePath` value in function arguments.

If we derived filePath from with in `Ariadne.Wallet.Cardano.Kernel.Keystore.Persistence` then it would be easy to get rid of filePath from with in `keystore`.

gromakovsky: Ok, I agree, I didn't notice that. However, instead of storing `FilePath` in `InternalStorage`, why not store it in `Keystore`? This way we won't have to set default value in JSON deserialization.

epicallan: That sounds good.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
epicallan: I changed the name accordingly.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/Keystore.hs
```diff
@@ -34,15 +35,24 @@ import Pos.Core (AddressHash, addressHash)
...
+  (InternalStorage(..), isWallets)
+import Ariadne.Wallet.Cardano.Kernel.Keystore.Util
+  (eskToPublicHash, walletIdToPublicHash)
 import Ariadne.Wallet.Cardano.Kernel.Types (WalletId(..))
 
-data Keystore = Keystore (MVar InternalStorage)
+data Storage
+  = Storage
+  { _sIstorage :: !InternalStorage
+  , _sFilePath :: !FilePath
```
gromakovsky: What's the purpose of putting `FilePath` into `MVar`? It's constant, why not make it a pure value?

epicallan: I didn't think through it deeply. I am going to change it such that I have this kind of constructor instead:

```
data Keystore
  = Keystore
  { _kstorage :: !(Mvar InternalStorage)
  , _sFilePath :: !FilePath
  }
```

PR #349, opened 2018-10-12
[AD-369] Call newPending in sendTx

Pastafarianist:
**Description:**

This is a duplicate of https://github.com/serokell/ariadne/pull/328, which was automatically closed upon merging https://github.com/serokell/ariadne/pull/321.

This PR reintroduces the `ActiveWallet` & `ActiveWalletLayer` and adds a call to `newPending` to `sendTx`. Design-wise, this goes the opposite way compared to the original implementation (the intent was to add a transaction to the submission layer after a successful call to `newPending`), but this is OK since semantics-wise we do the same thing: we first create a new transaction, then try calling `newPending`, and if the call fails, we propagate the exception without submitting the transaction.

The tests now also call `newPending`. This helped fix one of the two failing tests left after AD-167.

**YT issue:** https://issues.serokell.io/issue/AD-369

**Checklist:**

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
@@ -193,7 +193,9 @@ sendTx
                 ourAddresses
                 (map TxOutAux outs)
                 (const newChangeAddress)
-        liftIO $ whenLeftM (awlNewPending awl ourAccountId txAux) throwM
+        -- TODO: call newPending even when inputs are selected from multiple accounts.
+        when (length accountsToUse == 1) $
+            liftIO $ whenLeftM (awlNewPending awl ourAccountId txAux) throwM
         let tx = taTx txAux
         let txId = hash tx
         liftIO $ printAction $ formatSubmitTxMsg tx
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
@@ -193,7 +193,7 @@ sendTx
                 ourAddresses
                 (map TxOutAux outs)
                 (const newChangeAddress)
-        -- TODO: call newPending even when inputs are selected from multiple accounts.
+        -- [AD-518] TODO: call newPending even when inputs are selected from multiple accounts.
         when (length accountsToUse == 1) $
             liftIO $ whenLeftM (awlNewPending awl ourAccountId txAux) throwM
         let tx = taTx txAux
--- a/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
@@ -35,7 +35,8 @@ import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet
 import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet.Read
 import Ariadne.Wallet.Cardano.Kernel.DB.InDb
 import Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet (IxSet, (@+))
-import Ariadne.Wallet.Cardano.WalletLayer.Types (PassiveWalletLayer(..))
+import Ariadne.Wallet.Cardano.WalletLayer
+  (ActiveWalletLayer(..), PassiveWalletLayer(..))
 import Ariadne.Wallet.Face
 
 data SendTxException
@@ -71,9 +72,9 @@ instance Exception SendTxException where
 -- may be used as inputs.  If this list is empty and an account from
 -- the input wallet is selected, this account will be used as input.
 -- Otherwise inputs will be selected from all accounts in the wallet.
-sendTx ::
-       (HasConfigurations)
-    => PassiveWalletLayer IO
+sendTx
+    :: HasConfigurations
+    => ActiveWalletLayer IO
     -> WalletFace
     -> CardanoFace
     -> IORef (Maybe WalletSelection)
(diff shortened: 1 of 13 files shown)


ariadne/cardano/test/backend/Test/Infrastructure/Generator.hs
```diff
@@ -74,8 +74,7 @@ genInductiveUsingModel GeneratorModel{..} = do
       }
   where
     potentialOurs = filter gmPotentialOurs gmAllAddresses
-    -- TODO: newPending probability is set to 0 until AD-369 is resolved
-    params ours   = (defEventsParams gmEstimateFee gmAllAddresses ours initUtxo) {gepPendingProb = 0.0}
+    params ours   = (defEventsParams gmEstimateFee gmAllAddresses ours initUtxo)
```
maksbotan: brackets?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -117,6 +118,11 @@ instance Buildable NewPendingError where
     build (NewPendingFailed npf) =
         bprint ("NewPendingFailed " % build) npf
 
+instance Show NewPendingError where
+    show = formatToString build
+
+instance Exception NewPendingError
+
 newPending :: HdAccountId
```
maksbotan: What does this do?

ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -189,6 +192,9 @@ sendTx
                 ourAddresses
                 (map TxOutAux outs)
                 newChangeAddress
+        liftIO $ awlNewPending awl ourAccountId txAux >>= \case
```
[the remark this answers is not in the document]
gromakovsky: @maksbotan I suppose it should be `whenLeftM`. I'm not sure if it's more readable either.
What I find strange is that our rules do have a rule about it: https://github.com/serokell/ariadne/blob/867105ca014f623684993895bdb484e681f29e74/.hlint-universum.yaml#L207, but there's no warning here (tried `make lint` locally).

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel.hs
```diff
@@ -147,6 +157,41 @@ observableRollbackUseInTestsOnly :: PassiveWallet -> IO ()
...
+    -> ComponentM ActiveWallet
+activeWalletComponent walletPassive = do
+    let logMsg = _walletLogMessage walletPassive
+    buildComponent "Active wallet"
+        (return ActiveWallet{..})
+        (\_ -> liftIO $ do
+            logMsg Error "stopping the wallet submission layer..."
+        )
+
+bracketActiveWallet
```
gromakovsky: Didn't we decide to put such `bracket*` functions to tests, because they are used only there?

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
```diff
@@ -117,6 +118,11 @@ instance Buildable NewPendingError where
     build (NewPendingFailed npf) =
         bprint ("NewPendingFailed " % build) npf
 
+instance Show NewPendingError where
+    show = formatToString build
+
+instance Exception NewPendingError
```
gromakovsky: I think it makes more sense to have automatically defined `Show` (so that it corresponds to the exception type structure, ideally should be machine readable) and use `displayException` for pretty-printing.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/AcidState.hs
gromakovsky: If this comment is about `HdAccountId` argument, I have the same question. If not, I have a question about this argument :)
Does it mean that currently `newPending` assumes only one input account? If yes, then it's bad, because it's possible to send a transaction from multiple accounts in our case.

Pastafarianist: > Does it mean that currently newPending assumes only one input account?

Yes. This issue is a bug, and it is a bit hard to work around. Currently this `HdAccountId` is chosen as the account that owns the change address. The purpose of this parameter is to specify which account checkpoints to use for storing the pending transaction. One of the things that is done when inserting a new pending transaction is checking that all of its inputs are available in the account that the transaction belongs to.

Hence merging this PR in its current state will break sending transactions from multiple accounts.

gromakovsky: > Hence merging this PR in its current state will break sending transactions from multiple accounts.

Okay, then I think we should merge in its current state and there are two options:
1. Implement it properly in this PR.
2. Comment out _something_, so that `newPending` is still called, but does nothing. Create an issue to implement it properly (with proper description).

ariadne/cardano/src/Ariadne/Wallet/Backend/Tx.hs
```diff
@@ -193,7 +193,9 @@ sendTx
                 ourAddresses
                 (map TxOutAux outs)
                 (const newChangeAddress)
-        liftIO $ whenLeftM (awlNewPending awl ourAccountId txAux) throwM
+        -- TODO: call newPending even when inputs are selected from multiple accounts.
```
gromakovsky: > Create an issue to implement it properly (with proper description).

Please do it and leave issue ID here. I'll approve after that.

maksbotan: Still no issue ID here :thinking:

PR #355, opened 2018-10-22
Store cursor in the parse tree

artemohanjanyan:
**Description:** Cursor is now stored inside the tree. This allows to use all of the input in the autocompletion (before only text before the cursor was parsed for autocompletion purposes) and has potential use cases for other editing features. On the other hand, it is quite hard to manipulate the tree with all that formatting info, and maybe something should be done about it.

**YT issue:** https://issues.serokell.io/issue/AD-458

**Checklist:**

  - [x] [README](README.md)
  - [x] [GUI usage guide](docs/usage-gui.md)
  - [x] [GUI architecture documentation](docs/gui-architecture.md)

### DIFF
--- a/knit/src/Knit/Autocompletion.hs
+++ b/knit/src/Knit/Autocompletion.hs
@@ -2,7 +2,13 @@ module Knit.Autocompletion
        ( suggestions
        ) where
 
+import Control.Applicative
+import Control.Monad
+import Control.Monad.Trans
+import Control.Monad.Trans.State.Strict
+import Data.Default
 import Data.List (isPrefixOf)
+import qualified Data.List.NonEmpty as NonEmpty
 import Data.Loc (loc, spanFromTo)
 import Data.Maybe (mapMaybe)
 import Data.Semigroup ((<>))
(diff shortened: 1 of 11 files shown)


knit/src/Knit/Autocompletion.hs
```diff
@@ -32,209 +37,276 @@ suggestions
...
-    cmd' = "(" <> cmd
-    tokens = snd $ tokenize cmd'
+    (spaceBefore, tokens) = tokenize cmd
 
     tokenBalance = \case
       TokenParenthesis bs -> withBracketSide 1 (-1) bs
       _ -> 0
-    parensBalance = sum $ map (tokenBalance . _lItem) tokens
-    cmdLines = T.splitOn "\n" cmd'
+    parensBalance = sum $ map (tokenBalance . _lItem . _twsToken) tokens
```
int-index: I'm still unhappy about this parentheses juggling. Is there at least a task to remove it? It seems it would involve adding a new terminal `eof` to Earley, but I'd like to see it done.

int-index: I created https://github.com/ollef/Earley/issues/37

artemohanjanyan: - I don't see how `eof` is better than this approach.
- It is possible to do without `eof` from Earley, we can extend `Knit.Tokenizer.Token` with our own eof (either directly or via an internal component), but
- It will be gone after https://issues.serokell.io/issue/AD-449 anyway.

knit/src/Knit/Autocompletion.hs
```diff
@@ -32,209 +39,275 @@ suggestions
...
+                (LeftSpace $ makeSws spaceBefore))
+              (RightSpace $ makeRightSws spaceAfter)
+
+newtype LeftSpace = LeftSpace { getLeftSpace :: SpaceWithSelection }
+  deriving (Default)
+
+newtype RightSpace = RightSpace { getRightSpace :: SpaceWithSelection }
+  deriving (Default)
+
+type SuggestionMonad = StateT LeftSpace (StateT RightSpace [])
```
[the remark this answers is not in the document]
artemohanjanyan: Makes sense, will do.

knit/src/Knit/Autocompletion.hs
int-index: > I don't see how `eof` is better than this approach.

It doesn't modify the input before parsing. Here you artificially add parentheses and then remove them, for this you compute the location of the last token, and for this you have to do `T.splitOn "\n" cmd'`. It's a mess if you ask me.

> It is possible to do without eof from Earley, we can extend Knit.Tokenizer.Token with our own eof (either directly or via an internal component)

I'm not sure that'd work, would you generate an infinite sequence of `TokenEof` at the end or just one? You would need an amount equal to that of unbalanced parentheses, and at this point it's barely an improvement.

int-index: > It will be gone after https://issues.serokell.io/issue/AD-449 anyway.

Well, I think you'll need the `eof` functionality in Early to do AD-449.

knit/src/Knit/Autocompletion.hs
```diff
@@ -130,17 +55,16 @@ suggestionExprs procs = goExpr
       lit@(ExprLit _ _) -> pure lit
       ExprProcCall NoExt p -> ExprProcCall NoExt <$> goPc p
       XExpr (ExprInBrackets l e r) -> do
-        oldLeft <- liftLeft get
-        oldRight <- liftRight get
-        liftLeft $ put $ LeftSpace $ fbSpace l
-        liftRight $ put $ RightSpace $ fbSpace r
+        oldLeft <- use leftSpace
```
int-index: Is there reason to have `oldLeft` and `oldRight` separately as opposed to `oldSuggestionCtx <- get`?

artemohanjanyan: missed it

PR #350, opened 2018-10-15
[AD-382] Tool for viewing current content of acid-state

Haskell-mouse:
**Description:**

Added utility that displays acid state DB with hdwallets (and can display it's previous states).
See readme with examples of using it.

**YT issue:** https://issues.serokell.io/issue/AD-382

**Checklist:**

- [ ] Updated docs if necessary
  - [ ] [README](README.md)
  - [ ] [TUI usage guide](docs/usage-tui.md)
  - [ ] [Configuration documentation](docs/configuration.md)
- [ ] My code complies with the [style guide](docs/code-style.md)
- [ ] Tested my changes if they modify code

PR #379, opened 2018-11-19
[AD-502] `getstate` function fails if there is an empty db

Haskell-mouse:
**Description:**

This PR fixes the bugs that were added in the  PR #350 due to the bug in acid-state: 
  1. Ariadne failed to be started with an empty database.
  2. `openLocalState` function loads not the current state of DB, but the previous one.
Also, it adds tests to check the behavior of the `acid-state` related functions. 

**YT issue:** https://issues.serokell.io/issue/AD-452

**Checklist:**

  - [ ] [README](README.md)
  - [ ] [GUI usage guide](docs/usage-gui.md)
  - [ ] [GUI architecture documentation](docs/gui-architecture.md)

### DIFF
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
@@ -70,7 +70,7 @@ import qualified Data.ByteString as BS
 import qualified Data.IxSet.Typed as IxSet
 import Data.SafeCopy (base, deriveSafeCopySimple)
 
-import Test.QuickCheck (Arbitrary(..), elements, oneof, vectorOf)
+import Test.QuickCheck (Arbitrary(..), Gen, elements, oneof, vectorOf)
 
 import qualified Data.Text.Buildable
 import Formatting (bprint, build, (%))
@@ -95,7 +95,7 @@ newtype WalletName = WalletName
       deriving newtype (Buildable)
 
 instance Arbitrary WalletName where
-    arbitrary = pure "New wallet"
+    arbitrary = WalletName . toText <$> (arbitrary :: Gen String)
 
 -- | Account name
 newtype AccountName = AccountName
@@ -154,6 +154,7 @@ data HasSpendingPassword =
 
     -- | If there is a spending password, we record when it was last updated.
   | HasSpendingPassword (InDb Core.Timestamp)
+  deriving Eq
 
 instance Buildable HasSpendingPassword where
     build NoSpendingPassword = "no"
@@ -230,6 +231,7 @@ data HdRoot = HdRoot {
       -- | When was this wallet created?
     , _hdRootCreatedAt   :: InDb Core.Timestamp
     }
+    deriving Eq
 
 instance Buildable HdRoot where
     build HdRoot{..} =
@@ -259,6 +261,9 @@ data HdAccount = HdAccount {
       -- as stipulated by the wallet specification
     , _hdAccountCheckpoints :: NonEmpty AccCheckpoint
     }
+instance Eq HdAccount where
+  acc1 == acc2 =
+    _hdAccountId acc1 == _hdAccountId acc2 && _hdAccountName acc1 == _hdAccountName acc2
 
 instance Buildable HdAccount where
     build HdAccount{..} =
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
@@ -261,9 +261,7 @@ data HdAccount = HdAccount {
       -- as stipulated by the wallet specification
     , _hdAccountCheckpoints :: NonEmpty AccCheckpoint
     }
-instance Eq HdAccount where
-  acc1 == acc2 =
-    _hdAccountId acc1 == _hdAccountId acc2 && _hdAccountName acc1 == _hdAccountName acc2
+    deriving Eq
 
 instance Buildable HdAccount where
     build HdAccount{..} =
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
@@ -105,6 +105,9 @@ newtype IxSet a = WrapIxSet {
 instance Show a => Show (IxSet a) where
     show = show . map unwrapOrdByPrimKey . IxSet.toList . unwrapIxSet
 
+instance (Eq a, Indexable a) => Eq (IxSet a) where
+    (WrapIxSet ixSet1) == (WrapIxSet ixSet2) = ixSet1 == ixSet2
+
 -- | Evidence that the specified indices are in fact available
 type Indexable a = IxSet.Indexable (PrimKey a ': IndicesOf a) (OrdByPrimKey a)
 
--- a/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
+++ b/ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
@@ -101,12 +101,10 @@ type family IndicesOf (a :: *) :: [*]
 newtype IxSet a = WrapIxSet {
       unwrapIxSet :: IxSet.IxSet (PrimKey a ': IndicesOf a) (OrdByPrimKey a)
     }
+deriving instance Indexable a => Eq (IxSet a)
 
 instance Show a => Show (IxSet a) where
     show = show . map unwrapOrdByPrimKey . IxSet.toList . unwrapIxSet
-
-instance (Eq a, Indexable a) => Eq (IxSet a) where
-    (WrapIxSet ixSet1) == (WrapIxSet ixSet2) = ixSet1 == ixSet2
 
 -- | Evidence that the specified indices are in fact available
 type Indexable a = IxSet.Indexable (PrimKey a ': IndicesOf a) (OrdByPrimKey a)
--- a/ariadne/cardano/test/backend/Test/Spec/AcidState.hs
+++ b/ariadne/cardano/test/backend/Test/Spec/AcidState.hs
@@ -1,5 +1,8 @@
 module Test.Spec.AcidState (spec) where
 
+import Control.Exception.Base (ErrorCall(..), handle)
+import Control.Monad.Component hiding (throwM)
+-- import qualified Control.Monad.Component as C (throwM)
 import System.IO.Temp
 
 import Data.Acid (closeAcidState, openLocalStateFrom, query)
@@ -27,23 +30,38 @@ spec :: Spec
 spec = describe "Checking AcidState related functions behavior" $ do
   prop "Check empty DB state" $ withMaxSuccess 1 $
     monadicIO $ run $ withSystemTempDirectory "testWalletDBEmpty" $ \path -> do
-      getState path defDB 0 True >>= \case
-             Left err -> fail err
-             Right (acidDB, eventTag) -> do
-               db <- liftIO $ query acidDB Snapshot
-               let isEmptyDB = checkEmptyWalletDB db
-               return $ isEmptyDB && (eventTag == noEventsMsg)
+        getState path defDB 0 True >>= \case
+          Left _ -> return False
+          Right (acidDB, eventTag) -> do
+            db <- liftIO $ query acidDB Snapshot
+            let isEmptyDB = checkEmptyWalletDB db
+            return $ isEmptyDB && (eventTag == noEventsMsg)
   prop "Check that opened state is the same as it was before closing." $ withMaxSuccess 10 $ do
-            monadicIO $ do
-                passwds <- genSpendingPasswords 10
-                requests <- mapM genNewWalletRq $ ordNub passwds
-                pm <- pick arbitrary
-                liftIO $ withSystemTempDirectory "testWalletDBNonEmpty" $ \path -> do
-                  withLayerLocalStorage pm path $ \layer wallet -> do
-                    res <- checkAcidDBOpenedState wallet layer path requests
-                    case res of
-                      Left err -> fail $ toString err
-                      Right () -> pass
+    monadicIO $ do
+      passwds <- genSpendingPasswords 10
+      requests <- mapM genNewWalletRq $ ordNub passwds
+      pm <- pick arbitrary
+      liftIO $ withSystemTempDirectory "testWalletDBNonEmpty" $ \path -> do
+        handle emptyDBErrorHandler $
+          withLayerLocalStorage pm path $ \layer wallet -> do
+            res <- checkAcidDBOpenedState wallet layer path requests
+            case res of
+              Left err -> fail $ toString err
+              Right () -> pass
+  where
+    emptyDBErrorHandler :: ComponentError -> IO ()
+    emptyDBErrorHandler
+      (ComponentBuildFailed [ComponentAllocationFailed desc internalException] _) =
+        case fromException internalException of
+          Just (ErrorCall msg) ->
+            if desc == "Temp Storage DB" &&
+              msg == "getState returned Left with pos == 0"
+            then fail $ "Failed to open an empty acid-state DB,\
+              \ probably your version of acid-state is broken.\
+              \ The text of the internal error: " ++ msg
+            else throwM internalException
+          Nothing -> throwM internalException
+    emptyDBErrorHandler exception = throwM exception
 
 checkEmptyWalletDB :: DB -> Bool
 checkEmptyWalletDB (DB (HdWallets hdWallets hdAccounts hdAddresses)) =
@@ -65,7 +83,6 @@ checkAcidDBOpenedState pw pwl tempDBDir walletsToCreate = do
   closeAcidState oldDB
   newDB <- openLocalStateFrom tempDBDir defDB
   (DB (HdWallets newWallets _ _)) <- query newDB Snapshot
-
   if (oldWallets /= newWallets)
   then return . Left $ showDifferences oldWallets newWallets
   else return $ Right ()
--- a/ariadne/cardano/test/backend/Test/Spec/AcidState.hs
+++ b/ariadne/cardano/test/backend/Test/Spec/AcidState.hs
@@ -31,7 +31,7 @@ spec = describe "Checking AcidState related functions behavior" $ do
              Left err -> fail err
              Right (acidDB, eventTag) -> do
                db <- liftIO $ query acidDB Snapshot
-               let isEmptyDB = chekEmptyWalletDB db
+               let isEmptyDB = checkEmptyWalletDB db
                return $ isEmptyDB && (eventTag == noEventsMsg)
   prop "Check that opened state is the same as it was before closing." $ withMaxSuccess 10 $ do
             monadicIO $ do
@@ -45,8 +45,8 @@ spec = describe "Checking AcidState related functions behavior" $ do
                       Left err -> fail $ toString err
                       Right () -> pass
 
-chekEmptyWalletDB :: DB -> Bool
-chekEmptyWalletDB (DB (HdWallets hdWallets hdAccounts hdAddresses)) =
+checkEmptyWalletDB :: DB -> Bool
+checkEmptyWalletDB (DB (HdWallets hdWallets hdAccounts hdAddresses)) =
   all (== 0) [size hdWallets, size hdAccounts, size hdAddresses]
 
 noEventsMsg :: BL.ByteString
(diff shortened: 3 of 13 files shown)


ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+                requests <- mapM genNewWalletRq $ ordNub passwds
+                pm <- pick arbitrary
+                liftIO $ withSystemTempDirectory "testWalletDBNonEmpty" $ \path -> do
+                  withLayerLocalStorage pm path $ \layer wallet -> do
+                    res <- checkAcidDBOpenedState wallet layer path requests
+                    case res of
+                      Left err -> fail $ toString err
+                      Right () -> pass
+
+chekEmptyWalletDB :: DB -> Bool
```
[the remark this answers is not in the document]
Haskell-mouse: fixed and rebased to the current master.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+import Test.Spec.Fixture (genSpendingPasswords, withLayerLocalStorage)
+
+import Ariadne.Wallet.Cardano.Kernel.DB.AcidState (DB(..), Snapshot(..), defDB)
+import Ariadne.Wallet.Cardano.Kernel.DB.HdWallet (HdWallets(..))
+import Ariadne.Wallet.Cardano.Kernel.DB.Util.IxSet (IxSet, size)
+import Ariadne.Wallet.Cardano.Kernel.Internal (PassiveWallet(..), wallets)
+import Ariadne.Wallet.Cardano.WalletLayer (PassiveWalletLayer, pwlCreateWallet)
+
+spec :: Spec
+spec = describe "Checking AcidState related functions behavior" $ do
```
artemohanjanyan:
```suggestion
spec = describe "AcidState" $ do
```

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+  prop "Check empty DB state" $ withMaxSuccess 1 $
```
artemohanjanyan:
```suggestion
  prop "can open empty database" $ withMaxSuccess 1 $
```

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+    monadicIO $ run $ withSystemTempDirectory "testWalletDBEmpty" $ \path -> do
+      getState path defDB 0 True >>= \case
+             Left err -> fail err
+             Right (acidDB, eventTag) -> do
+               db <- liftIO $ query acidDB Snapshot
+               let isEmptyDB = checkEmptyWalletDB db
+               return $ isEmptyDB && (eventTag == noEventsMsg)
+  prop "Check that opened state is the same as it was before closing." $ withMaxSuccess 10 $ do
```
artemohanjanyan:
```suggestion
  prop "opened state is the same as it was before closing" $ withMaxSuccess 10 $ do
```

artemohanjanyan: I don't really understand this test. I reverted `stack.yaml` to the commit before your fixes and this test fails, but it fails somewhere inside acid-state rather than because of the checks you are doing here.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
artemohanjanyan: My options are consistent with other descriptions of the tests.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
Haskell-mouse: In this test, I create <=10 wallets, save the list of that wallets and close the acid-state DB. 
Then I open it again, get a list of wallets and compare with the previously saved list. 
If there are any differences then the test fails.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
Haskell-mouse: Ok, right. Thanks.

artemohanjanyan:
@gromakovsky Maybe we can merge it now and fix everything later with another PR?

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
Haskell-mouse: Added a handler for this exception.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,97 @@
+module Test.Spec.AcidState (spec) where
+
+import Control.Exception.Base (ErrorCall(..), handle)
+import Control.Monad.Component hiding (throwM)
+-- import qualified Control.Monad.Component as C (throwM)
```
artemohanjanyan: @Haskell-mouse looks like you committed comment accidentally.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,97 @@
...
+
+checkAcidDBOpenedState
+  :: PassiveWallet
+  -> PassiveWalletLayer IO
+  -> FilePath
+  -> [NewWallet]
+  -> IO (Either Text ())
+checkAcidDBOpenedState pw pwl tempDBDir walletsToCreate = do
+  let oldDB = pw ^. wallets
+  mapM_ (applyNewWallet  (pwlCreateWallet pwl)) walletsToCreate
```
artemohanjanyan: extra space

gromakovsky:
> Maybe we can merge it now and fix everything later with another PR?

Fix what? I am trying to understand now…

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
```diff
@@ -58,14 +76,18 @@ withPassiveWalletFixture :: MonadIO m
...
             fixtures <- generateFixtures wallet
             cc keystore layer wallet fixtures
 
 inMemoryDBComponent
     :: ComponentM (AcidState Kernel.DB)
 inMemoryDBComponent = buildComponent_ "In-memory DB" (openMemoryState defDB)
 
+tempDBComponent
+    :: FilePath -> ComponentM (AcidState Kernel.DB)
+tempDBComponent path = buildComponent_ "Temp Storage DB" (openLocalStateFrom path defDB)
```
gromakovsky: It's not really essential in tests, but I think a good style would be to use `closeAcidState` here.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
```diff
@@ -39,13 +43,27 @@ genSpendingPassword :: PropertyM IO PassPhrase
...
+
+withLayerLocalStorage
+  :: MonadIO m
+  => ProtocolMagic
+  -> FilePath
+  -> (PassiveWalletLayer m -> Kernel.PassiveWallet -> IO a)
+  -> m a
+withLayerLocalStorage pm pathToDB cc =
+    liftIO $ Keystore.bracketTestKeystore $ \keystore -> do
+        bracketKernelPassiveWallet pm devNull keystore (Filesystem pathToDB) $ \layer wallet -> do
```
gromakovsky: I don't like this copy-paste. Why not make a helper function with type `MonadIO m => ProtocolMagic -> FileOrMemoryDB -> (PassiveWalletLayer m -> Kernel.PassiveWallet -> IO a) -> m a`? And then `withLayerLocalStorage` and `withLayer` can use it (feel free to rename smth).

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
```diff
@@ -39,13 +43,27 @@ genSpendingPassword :: PropertyM IO PassPhrase
 genSpendingPassword =
     pick (frequency [(20, pure emptyPassphrase), (80, arbitrary)])
 
+genSpendingPasswords :: Int -> PropertyM IO [PassPhrase]
+genSpendingPasswords n = pick $ vector n
```
gromakovsky: So it uses `instance Arbitrary PassPhrase`. Logic there differs from `genSpendingPassword`.
It looks weird to me that `genSpendingPassword` and `genSpendingPasswords` use different logic to generate password.
I would expect this function to be defined as `genSpendingPasswords n = replicateM n genSpendingPassword`.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,97 @@
...
+spec :: Spec
+spec = describe "AcidState" $ do
+  prop "can open empty database" $ withMaxSuccess 1 $
+    monadicIO $ run $ withSystemTempDirectory "testWalletDBEmpty" $ \path -> do
+        getState path defDB 0 True >>= \case
+          Left _ -> return False
+          Right (acidDB, eventTag) -> do
+            db <- liftIO $ query acidDB Snapshot
+            let isEmptyDB = checkEmptyWalletDB db
+            return $ isEmptyDB && (eventTag == noEventsMsg)
```
gromakovsky: Well, this test is too low-level, essentially it tests internals of `acid-state`. I think it's better to use `withLayerLocalStorage` like you do in the second test. `withLayerLocalStorage` calls `openLocalStateFrom` which is actually used in production, while `getState` is some internal function which should be used directly only in very rare cases.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,97 @@
...
+  prop "opened state is the same as it was before closing" $ withMaxSuccess 10 $ do
```
gromakovsky: This test is not a regression test for the bug and it wasn't requested in this issue. Please don't do more work than necessary.
Of course I am not asking to delete it, because it's already written and seems to work. This comment applies to further work.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
gromakovsky: Also consider defining `walletDBComponent :: FilePath -> ComponentM (AcidState Kernel.DB)` in `Ariadne.Wallet.Cardano.WalletLayer.Kernel`, using it in `passiveWalletLayerComponent` and exporting solely for tests.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
```diff
@@ -259,6 +261,9 @@ data HdAccount = HdAccount {
       -- as stipulated by the wallet specification
     , _hdAccountCheckpoints :: NonEmpty AccCheckpoint
     }
+instance Eq HdAccount where
```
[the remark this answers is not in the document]
Haskell-mouse: Could we have 2 accounts with the same name, ID and with different checkpoints?

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
Haskell-mouse: `genSpendingPassword` for some reasons generate 20% of empty passphrases, but I need to have a list of unique. I even filter the list of random values. So - the cause of using different logic is that I don't want to generate 2 empty passwords and filter them every time. 

I will ask @Pastafarianist if I can remove that 20% of empty passphrases.

Haskell-mouse: I need unique passphrases because with the same passphrases I get the same requests and error when I try to create a wallet with the same ID.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/Util/IxSet.hs
```diff
@@ -105,6 +105,9 @@ newtype IxSet a = WrapIxSet {
 instance Show a => Show (IxSet a) where
     show = show . map unwrapOrdByPrimKey . IxSet.toList . unwrapIxSet
 
+instance (Eq a, Indexable a) => Eq (IxSet a) where
```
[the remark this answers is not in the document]
Haskell-mouse: Oh, It can be derived. I did not know about the standalone deriving instance, that is why I did it by hands. I 've just read [docs about it](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#stand-alone-deriving-declarations) and derived the instance instead and removed written by hands.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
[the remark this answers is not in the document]
Haskell-mouse: Anyway added comparison by checkpoints too.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,97 @@
+module Test.Spec.AcidState (spec) where
+
+import Control.Exception.Base (ErrorCall(..), handle)
+import Control.Monad.Component hiding (throwM)
+
+import System.IO.Temp
```
epicallan: Make the import explicit.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
gromakovsky: I don't get it.
1. `instance Arbitrary PassPhrase` is defined as follows:
  ```haskell
  instance Arbitrary PassPhrase where
      arbitrary = oneof [
          pure mempty,
          ByteArray.pack <$> vector 32
          ]
  ```
  So it generates empty passphrase with 50% probability, which is even greater than probability of generating empty passphrase in `genSpendingPassword`.
2.
> I need unique passphrases because with the same passphrases I get the same requests and error when I try to create a wallet with the same ID.

Why do you get the same requests? `genNewWalletRq` picks arbitrary `eskNoPass`. It doesn't depend on passphrase at all. If you get the same requests, it means that `eskNoPass      <- pick arbitrary` generates the same wallet IDs, which is unrelated to generation of passphrases.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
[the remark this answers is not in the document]
gromakovsky: > Could we have 2 accounts with the same name, ID and with different checkpoints?

We can't have them in one DB, but in general we can. For instance, if we want to write a deserialization test (`decode . encode`), we should compare all fields. So I'd rather not assume where two `HdAccount`s come from. And if you want to assume that accounts are from the same DB, then comparing `HdAccountId`s should be enough.

Apart from that look at `HdAddress`, its `Eq` instance is derived and compares everything, so I'd like to be consistent with that.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
Haskell-mouse: 1. Oh! thanks! I thought that I get errors because random values can be duplicated and that `Arbitrary` instance should return really random value by default. 
Anyway, I'll add function to get a really random list of passphrases (I don't want to change the instance because I believe that there were reasons to get 50% of empty passphrases). 
(I thought "why I get so many duplications ??")

And yes, you are right - I'll use `genSpendingPassword` function.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
Haskell-mouse: Ok.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
Haskell-mouse: Ok. done.

ariadne/cardano/src/Ariadne/Wallet/Cardano/Kernel/DB/HdWallet.hs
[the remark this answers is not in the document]
Haskell-mouse: Ok, get it. thanks.

ariadne/cardano/test/backend/Test/Spec/Fixture.hs
Haskell-mouse: > It doesn't depend on passphrase at all. 

It is strange because with unique passphrases it actually generates different wallets. 
If i remove `OrdNub` from ```requests <- mapM genNewWalletRq $ ordNub passwds``` it generates sometimes duplicated requests (and duplicated wallets as well).

stack.yaml
```diff
@@ -102,7 +102,7 @@ extra-deps:
 # Because it includes a bunch of safecopy instances that we currently rely upon.
 # Also the 'Serokell.AcidState.ExtendedState' thing.
 - git: https://github.com/serokell/acid-state
-  commit: 219c7217ef52318a74bfa2aa54006b2505112f83
+  commit: 77291b5f7796a919414ce86fa47733e767bc5fdf
```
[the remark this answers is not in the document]
gromakovsky: `acid-state` PR has been merged, so you can do it.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
Haskell-mouse: Ok. I know that the first test is enough to detect this specific bug, I thought that it would be good to detect both manifestations of the bug. It can help in future ). 

Anyway:

gromakovsky: This bug is a showstopper or at least critical, it should be fixed ASAP. If the regression test was hard to write I would even allow you to merge this PR without addition of any tests, but in this particular case the regression test is quite simple, that's why I want it to be done together with the fix (but not other more complex tests).

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
Haskell-mouse: Ok. Done.

ariadne/cardano/test/backend/Test/Spec/AcidState.hs
```diff
@@ -0,0 +1,103 @@
...
+import Ariadne.Wallet.Cardano.Kernel.Internal (PassiveWallet(..), wallets)
+import Ariadne.Wallet.Cardano.WalletLayer (PassiveWalletLayer, pwlCreateWallet)
+
+spec :: Spec
+spec = describe "AcidState" $ do
+  prop "can open empty database" $ withMaxSuccess 1 $
+    monadicIO $ do
+      pm <- pick arbitrary
+      liftIO $ withSystemTempDirectory "testWalletDBEmpty" $ \path -> do
+        handle emptyDBErrorCatch $
```
gromakovsky: Do we need it?

Haskell-mouse: I think yes.  
Otherwise we get only an exception from the second property (i think that exception is not a valid result of test). 
We can catch it in the second propety, but in that case we would not know why the test failed - because Ariadne can't open empty database or it loads acid db state in a wrong way.

Haskell-mouse: I thought about it yesterday and think that we should leave both peroperties

gromakovsky: I don't get it.
If `withLayerLocalStorage @IO pm path $ \_ wallet -> checkEmptyDB wallet` throws `ComponentError`, it will be caught and the result will be `False` which is the same as if `checkEmptyDB` returns `False`.
So we can't distinguish whether the DB couldn't be open or it was open with unexpected state, can we?

gromakovsky: OTOH if we don't catch anything here, in the first case we'll get an exception (which is good, because the test should fail anyway) with error message and in the second case we'll get `Falsifiable` which would mean that `checkEmptyDB` returned `False`.

gromakovsky: > Oh. I thought that you say about the whole first property.

No, I mean exactly this line which handles a particular exception.

I am going to approve this PR because it already took a lot of time and the bug is critical and the things we are discussing here are quite minor. However, it might make sense to open a follow-up PR.

gromakovsky: > I think that exception is a bad way to say that a test failed. I think that a test should return Testable that inform a programmer that a property is not satisfied anymore. An exception is a way to said that something went wrong during a test, and it is not valid or could not be done.

My opinion on that in a nutshell:
1. I don't see anything bad in having a test fail by means of exception. When you write a test and don't catch a certain exception it means that you expect that this exception won't be thrown. Testing framework takes care of handling exceptions that your code doesn't catch.
2. As I said, in this particular case the only change visible to the person running this test is that they no longer can distinguish failure due to wrong contents of DB and failure to open DB. Which is quite bad.
3. You could avoid (2) if instead of returning `False` you were using `fail` inside `PropertyM` monad. But the end result is almost the same as not catching the exception at all. The only benefit of doing so is that you could modify the error message. And the disadvantage is that it makes the test significantly larger and more complex.
4. We don't expect this exception to be thrown anymore. So it's essentially the same as any other exception. Any other exception can be thrown, but we don't catch them. We could catch all exceptions, but fortunately we don't have to, because it's done for us by the testing engine.

> I think now, that I should make 3 properties - opening DB, an emptiness of an empty DB and the equality of the closed and opened state.

I think it would be a move into wrong direction. First two tests would be doing basically the same thing: open empty DB. Instead of doing the same thing twice, results of the test can be made more specific. As I said, if you are strictly opposed to handling it by means of builtin runtime exceptions, you can do it using `fail` inside the `PropertyM` monad.

PR #358, opened 2018-10-23
[AD-432] Support custom finalizers for the UI

rvem:
**Description:** Backend exceptions couldn't be successfully handled by Qt app due to it's FFI calls. Now when exception on backed thread happens, it is being handled by custom link function, that finalize UI and rethrow this exception to UI thread.
To check this add something like `threadDelay 5000000 >> throwM (ExitFailure 42)` to [serviceAction](https://github.com/serokell/ariadne/blob/b21274c96038069f42293dacbe621c8e95432790/ariadne/cardano/src/Ariadne/MainTemplate.hs#L158)

**YT issue:** https://issues.serokell.io/issue/AD-432

**Checklist:**

- [x] Updated docs if necessary
  - [x] [README](README.md)
  - [x] [TUI usage guide](docs/usage-tui.md)
  - [x] [GUI usage guide](docs/usage-gui.md)
  - [x] [Configuration documentation](docs/configuration.md)
  - [x] [GUI architecture documentation](docs/gui-architecture.md)
- [x] My code complies with the [style guide](docs/code-style.md)
- [x] Tested my changes if they modify code

PR #397, opened 2018-12-04
[AD-436] Make common Face and Glue

Hithroc:
## Description

This PR generalizes `Glue.hs` and `Face.hs` over frontends. Now, instead of having two mostly the same Face and Glue for `qt` and for `vty`, we now have common Face and Glue modules both of which do not depend on any specific frontend, however, they can be extended by the frontend. Some of the ADTs in Face have special extension constructor which is defined by the frontend. If in the future some other ADTs need to be extended, but they don't have such constructor, it can be easily added.

Here's a general structure of Common.Face and <FrontendName>.Face (replace `<FrontendName>` with the name of the frontend, e. g. `Vty`):

```haskell
-- Common.Face
data UiEvent frontend
  = UiCommonEvent1
  | UiCommonEvent2
  | UiFrontendEvent (FrontendEvent frontend)

data family FrontendEvent frontend

-- <FrontendName>.Face
data <FrontendName>

data instance FrontendEvent <FrontendName>
  = FrontendSpecificEvent1
  | FrontendSpecificEvent2
```

Note: in this PR only Vty code base is ported to the generalized modules, Qt remains untouched and will be done in a separate PR.

## Related issue(s)

https://issues.serokell.io/issue/AD-436

- Tests
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts


ui/common-lib/package.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+library:
+  <<: *lib-common
+  ghc-options:
+    - -Wno-unused-top-binds
+  dependencies:
+    - Earley
+    - Hclip
+    - ariadne-core
+    - ariadne-cardano
```
gromakovsky: This dependency looks wrong.

ui/common-lib/src/Ariadne/UI/Common/Face.hs
```diff
@@ -0,0 +1,304 @@
+module Ariadne.UI.Common.Face where
```
gromakovsky: All exports must be explicit.

ui/common-lib/package.yaml
gromakovsky: I see that it's needed for common glue. However, `ariadne-vty-lib` must not depend on any Cardano stuff, because it's also used in Disciplina.
The most obvious way to avoid it is to have 2 common libs (one with common face and the other one with common glue). I find it a bit disappointing to have 2 libs with one module, but can't quickly come up with a better approach.

ui/qt-app/Glue.hs
```diff
@@ -39,347 +24,23 @@ import Ariadne.UX.PasswordManager
...
-putCardanoEventToUI :: UiFace -> CardanoEvent -> IO ()
-putCardanoEventToUI UiFace{..} ev =
-  whenJust (cardanoEventToUI ev) putUiEvent
+putCommonUiEvent :: QtFace -> UiEvent -> IO ()
+putCommonUiEvent QtFace{..} event = putUiEvent $ UiCommonEvent event
 
 -- Called when an exception in backend thread occurs
-putBackendErrorToUI :: UiFace -> SomeException -> IO ()
-putBackendErrorToUI UiFace{..} e = putUiEvent . UiBackendExceptionEvent $
+-- TODO: Why is this not a thing in Vty?
```
gromakovsky: If backend fails in TUI, it's ok to have the program crash with a message printed to console (which is the default exception handling mechanism). In the GUI case, however, we want to show a dialog with the error. See #358.

gromakovsky (changes requested):

ui/common-lib/package.yaml
hithroc: I see the problem. Is this also why app and lib are different packages?

hithroc:
I've updated the PR.
It is quite a bit different from what it was originally, here are some major points:

* Qt code base is not touched at all. It will be done in a separate PR.
* The generalized Face now uses data families to let different frontends add their own things.

Basically, we have
```haskell
-- Common.Face
data UiEvent frontend
  = SomeCommonEvent1
  | SomeCommonEvent2
  | UiFrontendEvent (FrontendEvent frontend)

data family FrontendEvent frontend

-- Vty.Face
data Vty

data instance FrontendEvent Vty
  = SomeVtySpecificEvent
```

This makes it easy enough to extend the generalized Face with stuff without changing the current code base much.

During review please check that all the parts from original Glue and Face ended up either in common or Vty Glue/Face in case I accidentally deleted something (pretty sure I didn't, but...).

gromakovsky:
PR description must not be empty. Also:
```
Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

If you don't set a checkmark (e. g. don't add a test for new functionality),
you must be able to justify that.
```

gromakovsky:
FYI [this PR](https://github.com/serokell/ariadne/pull/407/files) has some changes related to your work. E. g. `UiPasswordSent` and `UiConfirmDone` are deleted there.

ui/common-face/src/Ariadne/UI/Common/Face.hs
```diff
@@ -0,0 +1,398 @@
...
+  | UiRestoreWalletCommandResult UiRestoreWalletCommandResult
+  | UiNewAccountCommandResult UiNewAccountCommandResult
+  | UiNewAddressCommandResult UiNewAddressCommandResult
+  | UiBalanceCommandResult UiBalanceCommandResult
+  | UiTxHistoryCommandResult UiTxHistoryCommandResult
+  | UiFeeCommandResult UiFeeCommandResult
+  | UiRenameCommandResult UiRenameCommandResult
+  | UiRemoveCommandResult UiRemoveCommandResult
+  | UiExportCommandResult UiExportCommandResult
+  | UiFronendCommandResult (FrontendCommandResult frontend)
```
gromakovsky: Typo (`Fronend`).

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
```diff
@@ -74,349 +23,31 @@ data UiFeatures = UiFeatures
...
-  deriving (Eq, Show)
-
-data UiTxHistoryRow = UiTxHistoryRow
-  { uthrId :: Text
-  , uthrTotal :: Text
-  , uthrFrom :: [UiTxHistoryRowPart]
-  , uthrTo :: [UiTxHistoryRowPart]
+data instance UiCurrency Vty
+  = UiCurrency
+  { getUiCurrency :: (Maybe Text)
```
gromakovsky: Please add a comment explaining what `Nothing` means. And why the fact that it can be `Nothing` belongs to this type, not the place where it's used (`Maybe UiCurrency`).
(I don't know how it's used, so it's not obvious to me whether it's a good idea or not)

ui/common-glue/src/Ariadne/UI/Common/Glue.hs
```diff
@@ -0,0 +1,417 @@
...
+
+import qualified Control.Concurrent.Event as CE
+import Control.Exception (displayException)
+import Control.Lens (ix)
+import Data.Coerce
+import Data.Double.Conversion.Text (toFixed)
+import Data.Tree (Tree(..))
+import Data.Unique
+import qualified Data.Vector as V
+--import Data.Version (Version)
```
gromakovsky: Probably should be deleted.

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
gromakovsky: Also do we want to have something different in Qt?

gromakovsky (changes requested):
LGTM, but I have some relatively minor comments.

hithroc:
>PR description must not be empty

Oh yeah, forgot to fill that in since creation, my bad.

>FYI this PR has some changes related to your work. E. g. UiPasswordSent and UiConfirmDone are deleted there.

That should be very easy to merge, removing `UiPasswordSent` and `UiConfrimDone` in the common Glue should be enough.

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
gromakovsky: > I wasn't actually sure what Maybe is for here, it was in the original Face.hs so I left it as is. But as far as I can see, this value is actually never can be Nothing.

I suspect it might be needed in Disciplina. Could you please check? I am even quite sure this `Maybe` is for Disciplina, but whether it (`Maybe`) should be part of this type or applied externally depends on how it's used in Disciplina.

goodsoft: My idea was for each backend to provide some kind of string identifier (`ada`, `dscp`, ...), which would later be used by a frontend to find a corresponding icon in resources, or use fallback text display otherwise.

This way, frontend normally doesn't care about what backend is actually used, but if it actually has some backend-specific resources bundled (e.g., currency icon), it can utilize them.

hithroc: Yeah, I believe a system like this is what I'm going to work on after I finish this task.

goodsoft:
I would really like to see Qt frontend ported to this common code as well before approving.
Otherwise, looks like simple moving around of code for now.

Also, I hope that this abundance of packages is temporary.
Ideally, I would even merge `vty-app` and `vty-lib` (correspondingly for `qt-*`) and have one `ariadne-ui` package with common/glue code.
The idea for splitting packages was to make different executables for different currencies, but the direction has changed to including multiple backends in one executable, so there's no point in splitting them from corresponding UI libraries anymore.

hithroc:
Didn't @gromakovsky just say earlier that the packages are split so `vty-lib` doesn't have any cardano dependencies?

goodsoft:
Yeah, right, disregard my comment.
Still, a package per module is too much, I think.

gromakovsky:
@hithroc did you start porting Qt frontend? I like that this PR doesn't have Qt changes, I think it makes sense to make a new PR with Qt changes so that @goodsoft can review them.

> the direction has changed to including multiple backends in one executable, so there's no point in splitting them from corresponding UI libraries anymore.

Well, in this sense I think you can treat this abundance of packages as temporary, but I doubt we'll manage to include multiple backends in one executable any time soon.

ui/vty-lib/src/Ariadne/UI/Vty/Face.hs
gromakovsky: Well, my primary concern is that it should be possible to use this new code in Disciplina.

hithroc:
Yes, I did start, in fact they were already in progress when this PR was created, I intentionally not included them in this PR (as stated in the description).

hithroc:
I also have checked Disciplina's code and it does in fact use that `Maybe`, but not exactly sure what for. So I guess I should revert that commit?

goodsoft:
Well, the problem is, I can't judge whether the changes in this PR make sense, until I see that Qt frontend works with them without any hacky workarounds or major changes in the common code.
Thus, I would try to put them in the same PR, but as a separate commit.

gromakovsky:
>  Thus, I would try to put them in the same PR, but as a separate commit.

The advantage of having these changes in separate PRs is that we might be able to merge the first part sooner (if some changes are requested regarding the second part which don't require changing the first one). It's good because others will be working with a newer version of the code ⇒ less conflicts.

> I also have checked Disciplina's code and it does in fact use that Maybe, but not exactly sure what for. So I guess I should revert that commit?

AFAIU, you have figured that out in Slack.

goodsoft:
Okay, how about a new PR, which depends on this one?

gromakovsky:
Yep, that's what I meant.

ui/vty-lib/src/Ariadne/UI/Vty/Widget/Wallet.hs
```diff
@@ -346,17 +346,7 @@ handleWalletWidgetEvent = \case
                     WalletAccount idx (fromMaybe "" uaciLabel) uaciBalance False)
                 (zip [0..] uwiAccounts)
           updateEditable walletAccountsL walletAccountsEditL converted
-          case uwiBalance of
-            Just balance ->
-              walletBalanceL .= BalanceResultSuccess balance
-            Nothing -> do
```
gromakovsky: AFAIU, this logic is still used in DSCP, will it be possible to migrate DSCP to this version without touching this code?

gromakovsky (commented):
LGTM, but I am concerned that some deleted logic might actually be useful, so I'd like to be sure it's not.

gromakovsky:
Also I reviewed last commit from #410, looks convincing, though I didn't read very carefully yet.

PR #398, opened 2018-12-05
[AD-519] Add proper logging

gromakovsky:
## Description

Problem: it's not clear what one should do when they want to log something from Ariadne code. One approach is to use log-warper's functions, because log-warper is used by cardano-sl. However, it's not clear from the code that it should be used.
Moreover, I think using log-warper is not the best idea (see below).
We can't just use `putStrLn`, because it will break TUI. Also it would be good to print additional data with each message. `writeFile` won't break TUI, but it's not clear to which file one should log.

Solution: add `Ariadne.Logging` module with a simple `co-log`-based logging implementation to `ariadne-core`.
Current logging behavior is quite simple: each message is logged into a file called `ariadne.log`, location of this file is configurable.
It would be good to also put messages into Logs widgets in UI.
In GUI we can also print them to terminal.
Some additional information is attached to each message, specifically: severity and call site. In order to make the latter better, you should add `HasCallStack` constraint to each place from which you log. But it's not necessary.
It would be good to also print at least timestamps.
`co-log` has been chosen because it's quite simple and we don't need too many features, at least for now. Also it's quite extensible, so we can have more features later.
A consequence of using `co-log` is that Cardano uses different logging engine. One alternative would be to modify Cardano code, but it would be too painful. Another alternative would be to use `log-warper`, but IOHK, to the best of my knowledge, are going to stop using `log-warper` (maybe they've done it already) and in Serokell we think that `log-warper` has several drawbacks and we are not going to maintain it. Overall I think this consequence is not a big deal.
Logging is implemented as an opaque handle which is passed to logging functions. No monadic interface is used for it. It's in line with our existing code base, which mostly operates in `IO` and uses explicit argument passing.
Another concern is concurrency. Currently it's possible that logging output will be messy if several threads use it concurrently. A simple solution for it is using an `MVar`. I haven't implemented it, because Ariadne itself is mostly single-threaded, so I don't expect it to be a
problem. We can do it later if it turns out to be a problem.

P. S. As you can see, there are several `Would be good` which I haven't done in this PR. The goal of AD-519 is to just make it easy to log something and I believe this PR achieves it. Further improvements can be done in follow-up issues.

## Related issue(s)

https://issues.serokell.io/issue/AD-519

- Tests
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts

### DIFF
--- a/ariadne/cardano/src/Ariadne/Config/CLI.hs
+++ b/ariadne/cardano/src/Ariadne/Config/CLI.hs
@@ -32,10 +32,12 @@ import System.Directory
 import System.FilePath (isAbsolute, takeDirectory, (</>))
 
 import Ariadne.Config.Ariadne
-  (AriadneConfig(..), acCardanoL, acHistoryL, acWalletL, defaultAriadneConfig)
+  (AriadneConfig(..), acCardanoL, acHistoryL, acLoggingL, acWalletL,
+  defaultAriadneConfig)
 import Ariadne.Config.Cardano
 import Ariadne.Config.DhallUtil (fromDhall)
 import Ariadne.Config.History
+import Ariadne.Config.Logging
 import Ariadne.Config.Update
 import Ariadne.Config.Wallet
   (WalletConfig(..), walletFieldModifier, wcAcidDBPathL, wcKeyfilePathL)
@@ -48,6 +50,7 @@ data CLI_AriadneConfig = CLI_AriadneConfig
     , cli_acWallet :: !CLI_WalletConfig
     , cli_acUpdate :: !CLI_UpdateConfig
     , cli_acHistory :: !CLI_HistoryConfig
+    , cli_acLogging :: !CLI_LoggingConfig
     } deriving (Eq, Show, Generic)
 
 data CLI_CardanoConfig = CLI_CardanoConfig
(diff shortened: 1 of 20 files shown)


gromakovsky:
If you are from Serokell, consider reading https://www.notion.so/serokell/Logging-3d5d8b1e800e4de6b10a5187efdd6848 as well.

ariadne/cardano/src/Ariadne/Config/CLI.hs
```diff
@@ -48,6 +50,7 @@ data CLI_AriadneConfig = CLI_AriadneConfig
     , cli_acWallet :: !CLI_WalletConfig
     , cli_acUpdate :: !CLI_UpdateConfig
     , cli_acHistory :: !CLI_HistoryConfig
+    , cli_acLogging :: !CLI_LoggingConfig
```
artemohanjanyan: Why do we have this copy-paste?

artemohanjanyan: (I mean most of this file)

gromakovsky: Because nobody knows how to do better.

artemohanjanyan: I don't understand the problem solved by this copy-paste.

gromakovsky: Because we have a configuration stored in file(s) and we want each field there to be overridable via CLI. For each mandatory field in the configuration we want a corresponding optional option in CLI.

artemohanjanyan: Hmm, ok. It was hard to notice any difference between the types.

PR #407, opened 2019-01-10
[AD-445] Do not use UiFace to disable modality

gromakovsky:
## Description

Problem: sometimes vty app can go into modal mode and some widgets use
`putUiEvent` to let the app know that it should exit modal mode. They do it
from the UI thread and because of that `putUiEvent` can block forever.
Solution: add a new event to `WidgetEvent` to disable modality and send this
event to parent using `widgetEvent`.

## Related issue(s)

https://issues.serokell.io/issue/AD-445

- Tests
  - [ ] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts

### DIFF
--- a/ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
+++ b/ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
@@ -173,6 +173,7 @@ data WidgetEvent
   | WidgetEventEditLocationChanged
   | WidgetEventListSelected Int
   | WidgetEventCheckboxToggled
+  | WidgetEventModalExited
 
 -- | Result of handling a UI event
 --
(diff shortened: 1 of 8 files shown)


ui/vty-lib/src/Ariadne/UI/Vty/Widget.hs
```diff
@@ -173,6 +173,7 @@ data WidgetEvent
   | WidgetEventEditLocationChanged
   | WidgetEventListSelected Int
   | WidgetEventCheckboxToggled
+  | WidgetEventNoModal
```
goodsoft: I would rather name it `ModalExited`, because the semantics of this ADT is "something happened"

gromakovsky:
I don't know either :)

PR #409, opened 2019-01-17
[AD-445] Do not send command to REPL from UI thread

gromakovsky:
## Description

Problem: `langPutUiCommand` is executed from the UI thread and it calls
`putUiEvent` which can block and only the UI thread can unblock it. So deadlock is possible.
Solution: call `putUiEvent` from Knit backend, inside action passed to
`spawnTask`. That action is executed from non-UI thread, so it can't deadlock.
One alternative would be to introduce a way to send an event from widgets to
REPL. However, that seems to be too hard.
Another alternative is to fork a thread specifically for `putUiEvent`, but
this event should be processed before the command is executed, because
otherwise `updateCommandResult` will ignore all events related to this command.
This problem was present before, because `putUiEvent` was called after
`putCommand`.
This approach is quite simple and doesn't require additional thread management.

This is the last fix for AD-445.

## Related issue(s)

https://issues.serokell.io/issue/AD-445

- Tests
  - [ ] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again. ← don't know how to do it.

- Documentation
    - Haddock

- Public contracts

### DIFF
--- a/ui/vty-app/Glue.hs
+++ b/ui/vty-app/Glue.hs
@@ -71,7 +71,7 @@ knitFaceToUI
   -> UiLangFace
 knitFaceToUI UiFace{..} KnitFace{..} putPass =
   UiLangFace
-    { langPutCommand = putCommand False Nothing
+    { langPutCommand = putCommand False False Nothing
     , langPutUiCommand = putUiCommand False
     , langPutUISilentCommand = putUiCommand True
     , langParse = Knit.parse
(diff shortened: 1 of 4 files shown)


pasqu4le:
LGTM too and I cannot think of another way to deal simply with the issue.

ui/vty-app/Glue.hs
```diff
@@ -82,26 +82,27 @@ knitFaceToUI UiFace{..} KnitFace{..} putPass =
...
-      fmap (commandIdToUI cid) . putKnitCommand (commandHandle silent mOp cid) $ expr
+      fmap (commandIdToUI cid) . putKnitCommand (commandHandle ui silent mOp cid) $ expr
 
     putUiCommand silent op = case opToExpr op of
       Left err -> return $ Left err
       Right expr -> do
         whenJust (extractPass op) pushPassword
-        comId <- putCommand silent (Just op) expr
-        unless silent $
-          putUiEvent . UiCommandEvent comId . UiCommandWidget $ Knit.ppExpr expr
```
maksbotan: How about removing this `UiCommandWidget` as well?

gromakovsky: How would you do it? We should notify REPL about it somehow. Notifying it directly is complicated, because it can happen in any widget and sending an event from arbitrary widget A to REPL is complicated.
Do you mean replacing it with something else?

